PR #5, opened 2017-04-21
Adding creator to voucher

mamccr:
@eanders I doubt this one is really done. All it's doing is setting up the data, but I wasn't sure if we wanted to display it anywhere. I also struggled with naming a bit on this one. I didn't follow convention for the Voucher -> User relationship, because I wanted to make it clear that the key referenced the creator of the voucher. I'm not sure if it's necessary or more/less confusing. We could easily just call it user and put a comment somewhere. Open to thoughts and feelings about that.


mamccr:
Is this what you had in mind, @eanders ?:

image

eanders:
Exactly.

PR #23, opened 2017-05-26
Stories/130310737 pump history into hmis

dfhoughton:
So I added pry-byebug for the sake of debugging -- it didn't seem there was any debugger yet in this project. The main thing, though, is the task code which populates the report table. See if that seems sane, please.

One thing I had to change from our discussion is that I filtered out "non-initialized" decisions, where these are decisions with nil statuses. It seemed there were a lot of these, but a match only has a current decision, according to the `current_decision` method, if it has initialized decisions, and these are decisions with some status. I figured maybe only initialized decisions were relevant, so I went ahead and filtered out all rows not corresponding to initialized decisions.

This isn't done, but I need to work on something else and hoped you'd have a look at the logic. The main thing that remains is to display the history on the HMIS side, which in principle is just a matter of choosing and arranging the data this code creates.

### DIFF
--- a/app/models/hmis/build_report.rb
+++ b/app/models/hmis/build_report.rb
@@ -1,0 +1,66 @@
+module Hmis
+  class BuildReport
+    def run!
+      # we just flush the whole damned thing every time
+      Hmis::Report.delete_all
+      at = MatchDecisions::Base.arel_table
+      ::Client.joins( :project_client, client_opportunity_matches: :decisions ).preload(
+        :project_client,
+        {
+          client_opportunity_matches: {
+            decisions: %i[
+              decline_reason
+              not_working_with_client_reason
+              administrative_cancel_reason
+            ]
+          }
+        }
+      ).where( at[:status].not_eq nil ).distinct.find_each do |client|
+        client_id = client.project_client.id_in_data_source
+        client.client_opportunity_matches.each do |match|
+          # similar to match.current_decision, but more efficient given that we've preloaded all the decisions
+          decisions = match.decisions.select(&:status).sort_by(&:created_at)
+          next if decisions.empty?
+          current_decision = decisions.last
+          previous_updated_at = nil
+          decisions.each_with_index do |decision, idx|
+            if previous_updated_at
+              elapsed_days = ( decision.updated_at.to_date - previous_updated_at.to_date ).to_i
+            end
+            Hmis::Report.create!(
+              client_id: client_id,
+              match_id: match.id,
+              decision_id: decision.id,
+              decision_order: idx,
+              match_step: decision.step_name,
+              decision_status: current_decision.label,
+              current_step: decision == current_decision,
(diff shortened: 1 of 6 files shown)


app/models/hmis/build_report.rb
```diff
@@ -0,0 +1,66 @@
+module Hmis
+  class BuildReport
+    def run!
+      # we just flush the whole damned thing every time
+      Hmis::Report.delete_all
```
eanders: Checkout the Utility class in the warehouse, let's implement the truncate method, it will be way faster for deletes.

app/models/hmis/build_report.rb
```diff
@@ -0,0 +1,66 @@
+module Hmis
```
eanders: Let's be consistent with the naming, as long and unwieldy as it is and call this module GrdaWarehouse instead of HMIS, which has another meaning in the homeless world and the warehouse.

app/models/hmis/build_report.rb
```diff
@@ -0,0 +1,66 @@
...
+  class BuildReport
+    def run!
+      # we just flush the whole damned thing every time
+      Hmis::Report.delete_all
+      at = MatchDecisions::Base.arel_table
```
eanders: I have a new pattern for these, not completely necessary, but I'm liking md_t (MatchDecision_Table) instead of the hard to determine which arel_table we're dealing with `at`.

app/models/hmis/build_report.rb
```diff
@@ -0,0 +1,66 @@
...
+              not_working_with_client_reason
+              administrative_cancel_reason
+            ]
+          }
+        }
+      ).where( at[:status].not_eq nil ).distinct.find_each do |client|
+        client_id = client.project_client.id_in_data_source
+        client.client_opportunity_matches.each do |match|
+          # similar to match.current_decision, but more efficient given that we've preloaded all the decisions
+          decisions = match.decisions.select(&:status).sort_by(&:created_at)
```
eanders: You probably checked this, but I'm not sure if you want to sort on created_at or id, or something else...

app/models/hmis/build_report.rb
```diff
@@ -0,0 +1,66 @@
...
+            end
+            Hmis::Report.create!(
+              client_id: client_id,
+              match_id: match.id,
+              decision_id: decision.id,
+              decision_order: idx,
+              match_step: decision.step_name,
+              decision_status: current_decision.label,
+              current_step: decision == current_decision,
+              decline_reason: explain( decision, :decline_reason ),
```
eanders: Nice!  I like having only one of these instead of needing the "other" too.

lib/tasks/hmis.rake
```diff
@@ -0,0 +1,10 @@
+namespace :hmis do
+  
+  desc "pump the latest client matches and decisions into the HMIS database"
```
eanders: maybe "pump all client matches..."

PR #27, opened 2017-06-14
Stories/130310737 pump history into hmis

eanders:


app/models/hmis/base.rb
```diff
@@ -0,0 +1,6 @@
+module Hmis
```
eanders: module should be Warehouse not Hmis

app/models/hmis/build_report.rb
```diff
@@ -0,0 +1,66 @@
...
+            if previous_updated_at
+              elapsed_days = ( decision.updated_at.to_date - previous_updated_at.to_date ).to_i
+            end
+            Hmis::Report.create!(
+              client_id: client_id,
+              match_id: match.id,
+              decision_id: decision.id,
+              decision_order: idx,
+              match_step: decision.step_name,
+              decision_status: current_decision.label,
```
eanders: Should this be decision.label?

app/models/hmis/report.rb
```diff
@@ -0,0 +1,6 @@
+# for writing latest state of matches X decisions into HMIS database
+module Hmis
+  class Report < Base
```
eanders: Let's call this Warehouse::CasReport

app/models/warehouse/build_report.rb
```diff
@@ -0,0 +1,66 @@
...
+              not_working_with_client_reason
+              administrative_cancel_reason
+            ]
+          }
+        }
+      ).where( at[:status].not_eq nil ).distinct.find_each do |client|
+        client_id = client.project_client.id_in_data_source
+        client.client_opportunity_matches.each do |match|
+          # similar to match.current_decision, but more efficient given that we've preloaded all the decisions
+          decisions = match.decisions.select(&:status).sort_by(&:created_at)
```
dfhoughton: This is not actually redundant. The query ignores clients that have only decisions with null statuses, but the preload then fetches in *all* decisions, so we pare them out here.

dfhoughton:
So I think this is ready to merge. I renamed some files and that obliterated the useful comments you had put there, but I think there weren't too many things to account for.

PR #29, opened 2017-07-24
Stories/translations interface 148903391

everestwitman:


app/controllers/admin/translation_keys_controller.rb
```diff
@@ -0,0 +1,94 @@
+module Admin
+  class TranslationKeysController < ApplicationController
+    # before_action :require_can_edit_translations!
```
eanders: Should this be commented out?

PR #47, opened 2017-10-25
Priority calculation

peterclark:


app/models/cas/update_clients.rb
```diff
@@ -56,6 +56,26 @@ def run!
               c.update_attributes(pc_attr)
```
eanders: Let's avoid the super efficient (and arguably easy to read) straight SQL update here.  We're already updating client on line 56, so we can use ruby to calculate the vispdat_priority_score and throw it on the fields to update.

peterclark: done.

PR #49, opened 2017-10-31
add last_seen rule

peterclark:


eanders:
@peterclark this will also need the field added to the update_clients task.

PR #93, opened 2018-05-02
Stories/form for deidentified client 156732902

everestwitman:


app/controllers/deidentified_clients_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  
+  def new
+    @deidentified_client = deidentified_client_source.new 
+  end
+  
+  def edit
+  end
+
+  def update
+    clean_params = clean_params(deidentified_client_params)
```
eanders: This needs to check the can change cohorts permission and remove the cohorts parameter if you can't.

app/controllers/deidentified_clients_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+    
+    def append_client_identifier dirty_params
+      dirty_params[:last_name] = "Anonymous - #{dirty_params[:client_identifier]}"
+      dirty_params[:first_name] = "Anonymous - #{dirty_params[:client_identifier]}"
+      dirty_params
+    end
+    
+    def reject_empty_cohort_ids dirty_params
+      dirty_params[:active_cohort_ids].reject! { |r| r.blank? }
```
eanders: @everestwitman Looks like create doesn't work if you don't have access to manage cohort ids on de-identified clients.  It will need a rescue or try here.

PR #96, opened 2018-05-09
two tabs for available and unavailable clients

everestwitman:


eanders:
I pulled this in manually since it needed some additional work to be compatible with the route-model branch.

Issue #13, opened 2017-05-15
Add additional attributes to units


eanders:
CAS Supports requirements at various levels of the vacancy tree, Sub-Grantee, Program, and Service, for example, but needs some more granularity at the unit level.  Specifically, some units have attributes that make them more appropriate for some clients.  We would like to support the following attributes on units:

- [ ] Ground Floor
- [ ] Wheelchair accessible
- [ ] Occupancy (variable 1-10)
- [ ] Household with Children
- [ ] Number of bedrooms
- [ ] Target population (Youth, Family, Individual)
 
Expose these attributes:

- [ ] On the unit edit page (this is where the ability add/remove requirements should be)
- [ ] On any match that includes it
- [ ] On the voucher lists that include the unit

Add supporting columns to `ProjectClient` and `Client`

- [ ] requires_ground_floor (boolean, default: false, null: false)
- [ ] requires_wheelchair_accessibility (boolean, default: false, null: false)
- [ ] required_number_of_bedrooms (integer, default: 1, null: false)

To accomplish the above, units need to `has_many :requirements`.  This should follow the same pattern of requirements for buildings.  The "attributes" listed above then become new rules that match the supporting columns added to `Client`.  Finally, any rules coming directly from the unit should show in a block on matches involving the unit in the top summary section, and on the voucher list page as noted.

eanders:
This should also influence the match rules, and get rolled-into an opportunity.  For example, If a unit is Wheelchair Accessible, the opportunity should require that the client need a wheelchair accessible unit.

PR #101, opened 2018-05-29
Add additional attributes to units

giverm:
Adds the following attributes to ```Unit```
- ground_floor
- wheelchair_accessible
- occupancy
- number_of_bedrooms
- target_population

Exposes attributes to views, adds  supporting columns to ```ProjectClient``` and ```Client```, and adds rules for matching on ground_floor, wheelchair_accessible, and number_of_bedrooms.

PR for issue: https://github.com/greenriver/boston-cas/issues/13

### DIFF
--- a/app/models/opportunity_details/via_voucher.rb
+++ b/app/models/opportunity_details/via_voucher.rb
@@ -7,7 +7,43 @@ module OpportunityDetails
     def unit_name
       voucher.unit.try :name
     end
-    
+
+    def unit_on_ground_floor?
+      return nil unless voucher.unit
+      voucher.unit.ground_floor? ? "Yes" : "No"
+    end
+
+    def unit_is_wheelchair_accessible?
+      return nil unless voucher.unit
+      voucher.unit.wheelchair_accessible? ? "Yes" : "No"
+    end
+
+    def unit_occupancy
+      return nil unless voucher.unit
+      voucher.unit.occupancy
+    end
+
+    def unit_bedrooms
+      return nil unless voucher.unit
+      voucher.unit.number_of_bedrooms
+    end
+
+    def unit_target_population
+      return nil unless voucher.unit
+      voucher.unit.target_population.titleize
+    end
+
+    def unit_details
+      return nil unless voucher.unit
+      [
+        "Ground Floor: #{unit_on_ground_floor?}",
+        "Wheelchair Accessible: #{unit_is_wheelchair_accessible?}",
+        "Occupancy: #{unit_occupancy}",
+        "Bedrooms: #{unit_bedrooms}",
+        "Target Population: #{unit_target_population}",
+      ]
+    end
+
     def building_name
       voucher.building.try :name
     end
@@ -21,11 +57,11 @@ module OpportunityDetails
       return nil if voucher.building.nil?
       voucher.building.map_link
     end
-    
+
     def program_name
       voucher.sub_program&.program&.name
     end
-    
+
     def sub_program_name
       voucher.sub_program&.name
     end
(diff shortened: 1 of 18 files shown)


giverm:
WIP

app/models/opportunity_details/via_voucher.rb
```diff
@@ -7,7 +7,43 @@ class ViaVoucher < Base
     def unit_name
       voucher.unit.try :name
     end
-    
+
+    def unit_on_ground_floor?
```
eanders: These should all be defined in the base class and return nil.  Since these require a unit they should be re-defined in `app/models/opportunity_details/via_unit.rb` where you won't need `return nil unless voucher.unit`

app/models/unit.rb
```diff
@@ -47,4 +52,8 @@ def self.associations_adding_requirements
   def self.associations_adding_services
     [:building]
   end
+
+  def self.target_population_select_array
```
eanders: Generally we'll name these something like `available_target_populations`  That way when it becomes a hash at a later date we don't have to rename it.

app/views/unit_for_building/_edit_form.html.haml
```diff
@@ -3,10 +3,15 @@
   .form-inputs.margin-top-02
     = f.input :available, as: :boolean
     = f.input :name, label: "#{Unit.model_name.human} name"
-        
+    = f.input :occupancy, collection: 1..10, include_blank: false
```
eanders: This seems small, but the collection should come from the unit `Unit.available_occupancies` which should also be used in the validation.  That way it all lines up if we decide to limit to 5 or increase to 15.  Same goes for the # of bedrooms below.

app/views/units/_form.html.haml
```diff
@@ -5,14 +5,18 @@
...
-        This unit is involved in a 
+      %p
+        This unit is involved in a
         = "#{link_to 'match', match_path(@unit.active_match)}, it's availability cannot be adjusted, nor can it be moved to another site.".html_safe
     %label.margin-bottom-00 Is the Unit Available?
     = f.input :available, disabled: disabled
     = f.association :building, label_method: :name, disabled: disabled, label: "#{Building.model_name.human}"
     = f.input :name
-        
+    = f.input :occupancy, collection: 1..10, include_blank: false
```
eanders: This seems like a great candidate for a partial.  This is the third copy of it.

db/migrate/20180528160945_add_attributes_to_units.rb
```diff
@@ -0,0 +1,9 @@
+class AddAttributesToUnits < ActiveRecord::Migration
+  def change
+    add_column :units, :ground_floor, :boolean
+    add_column :units, :wheelchair_accessible, :boolean
+    add_column :units, :occupancy, :integer, default: 1
+    add_column :units, :number_of_bedrooms, :integer
```
eanders: Let's default this to 1 as well.

db/rules.csv
```diff
@@ -53,4 +53,6 @@ VispdatScoreThreeOrLess,VI-SPDAT score of 3 or less,have
 VispdatScoreFourToSeven,VI-SPDAT score of 4 to 7,have
 VispdatScoreEightOrMore,VI-SPDAT score of 8 or more,have
 EnrolledInHmisProject,Enrolled in specified HMIS Project,be
-ActiveInCohort,Active in specified Cohort,be
\ No newline at end of file
+ActiveInCohort,Active in specified Cohort,be
+RequiringWheelchairAccessibility,Requiring wheelchair accessibility,be
```
eanders: These should read well. "Must be Requiring wheelchair accessibility" doesn't quite sound right to me.  How about "Must be wheelchair accessible", so this becomes `RequiringWheelchairAccessibility,Wheelchair accessible,be` and `RequiringGroundFloor,Ground floor accessible,be`

PR #113, opened 2018-08-13
Stories/unit requirements

nimashariatian:
as promised, here is another WIP for issue #13  
i specifically did not fully implement this all the way to get your feedback first. 

i also included the page object pattern for feature specs, so let me know what you think.
i have no problem reverting anything you don't find appealing :) 

also, do you think we can update the Rspec in the gems and add few more gems in the tests ?
i'm thinking capybara-screenshot . 

factory girl has also been renamed to factory bot, so we can take care of that as well.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -24,7 +24,7 @@
 #  * zeus: 'zeus rspec' (requires the server to be started separately)
 #  * 'just' rspec: 'rspec'
 
-guard :rspec, cmd: "bundle exec rspec" do
+guard :rspec, cmd: "bundle exec rspec --color" do
   require "guard/rspec/dsl"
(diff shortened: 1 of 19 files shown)


nimashariatian:
thanks for feedback @eanders :) 
i'll make the adjustments and continue working on this either on the evenings or on sunday

eanders:
@nimashariatian could you see if merging the `release-11` branch into your code now fixes the travis issues?  Thanks!

nimashariatian:
looks like specs passed

eanders:
Excellent.  We threw in some code to allow some tests to run without actually talking to the warehouse db, which isn't setup in the test environment.

PR #134, opened 2018-12-12
Filter, search, and sort for Non-HMIS clients

awisema:


eanders:
This looks generally good, but I'm having some issues with the UI.

1. I can't seem to sort and then filter (it throws an error)
screen shot 2018-12-13 at 7 54 55 pm
2. The list of agencies is odd and full of blanks.  (I have users with no agency - production will too).  I think `NonHmisClient#possible_agencies` isn't doing what you expect since you'll get back user.id and agency, not agency ID.  Agencies are just strings.
screen shot 2018-12-13 at 7 51 42 pm
3. Sort orders don't appear to match the selections.
screen shot 2018-12-13 at 7 52 26 pm
screen shot 2018-12-13 at 7 52 44 pm

eanders:
If time allows, we should implement what we have in the warehouse where the filter is subtly highlighted when there is an active filter.  (This should go elsewhere too, but it would be a good place to start.)
screen shot 2018-12-13 at 8 06 31 pm

PR #137, opened 2018-12-19
Story/add cas vacancies to warehouse 162739526

awisema:


app/models/voucher.rb
```diff
@@ -71,6 +71,22 @@ def self.associations_adding_services
     [:unit, :sub_program]
   end
 
+  def available_at
+    cursor = self
+
+    while cursor
+      if cursor.available
```
eanders: Doesn't this mean if the vacancy isn't currently available this will never catch?  As in, if this vacancy was created, made available, involved in a successful match and then made unavailable, this would return `nil`.  Perhaps, painful though it seems, we need to get all versions with the created_at and available and inspect that array.

app/models/warehouse/build_report.rb
```diff
@@ -4,6 +4,32 @@ def run!
...
+    end
+
+    def fill_cas_vacancy_table!
+      Warehouse::CasVacancy.delete_all
+
+      ::Voucher.all.each do |voucher|
+        vacancy = Warehouse::CasVacancy.new(program_id: voucher.sub_program.program.id, sub_program_id: voucher.sub_program_id)
+
+        if voucher.deleted_at
+          vacancy.destroy!
```
eanders: This shouldn't be necessary.  `vacancy` hasn't been persisted at this time and you only call `save!` in the else clause.  Perhaps, just `if voucher.deleted_at.blank?`

app/models/warehouse/build_report.rb
```diff
@@ -4,6 +4,32 @@ def run!
...
+          vacancy.sub_program_name = voucher.sub_program.name
+          vacancy.program_type = voucher.sub_program.program_type
+          vacancy.route_name = voucher.program.match_route.title
+
+          vacancy.vacancy_created_at = voucher.created_at
+          vacancy.vacancy_made_available_at = voucher.available_at
+          vacancy.first_matched_at = ClientOpportunityMatch.joins(:opportunity).
+              where(opportunities: { voucher_id: voucher.id } ).
+              order(Opportunity.arel_table[:created_at]).last&.created_at
+
```
eanders: `ClientOpportunityMatch.joins(:opportunity).
                where(opportunities: { voucher_id: voucher.id } ).maximum(Opportunity.arel_table[:created_at])` should work and will avoid instantiating the `ClientOpportunityMatch` object.

app/models/warehouse/cas_vacancy.rb
```diff
@@ -0,0 +1,6 @@
+module Warehouse
+  class CasVacancy < Base
+    belongs_to :program
```
eanders: These don't make a ton of sense since there are no programs or sub_programs in the warehouse database and these live there.  At the very least, you'll probably need to specify `class_name`.  If they aren't in use, I'd just remove the relationships (leaving the columns.)

app/models/voucher.rb
awisema: The `if` is detecting when the availability changes by checking if to see if the version at the cursor is available and the in the previous version it wasn't. If the version at the cursor isn't available, the cursor moves back to the previous version so that all of the history is inspected.

eanders: Right.  My mistake

PR #147, opened 2018-12-31
Story/routes need run engine flag 162742579

awisema:


app/jobs/matching/run_engine_job.rb
```diff
@@ -4,12 +4,12 @@ class RunEngineJob < BaseJob
     def perform
       # prevent the engine from running simultaneously
       ClientOpportunityMatch.with_advisory_lock(:engine_lock) do
-        MatchRoutes::Base.all_routes.each do |route|
+        # TODO replaced all_routes with available -- this means inactive routes will not be run, I believe this was the intended behavior, anyway?
+        MatchRoutes::Base.available.each do |route|
```
eanders: should be a safe assumption

app/models/matching/engine.rb
```diff
@@ -71,7 +72,7 @@ def create_candidate_matches opportunity
       }
       match =
         client.candidate_matches.create(opportunity: opportunity, client: client, universe_state: universe_state)
-      match.activate! if client_priority == 1
+      match.activate! if client_priority == 1 && activate_priority_match
```
eanders: I believe you don't need to pass `activate_priority_match` all the way down.  You should be able to get it from `opportunity.match_route.should_activate_match` or `match.match_route.should_activate_match`.  I believe that would minimize the changes here.

app/models/matching/engine.rb
```diff
@@ -5,7 +5,8 @@ def for_available_clients(opportunities, match_route: )
     end
 
     def create_candidates match_route:
-      new(available_clients(match_route: match_route), available_opportunities).replace_candidates
+      activate_priority_match = match_route.should_activate_match
+      new(available_clients(match_route: match_route), available_opportunities).replace_candidates(activate_priority_match)
```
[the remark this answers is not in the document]
eanders: If the comment below about not needing this variable at all is true, you can ignore the above.

PR #148, opened 2018-12-31
Story/new prioritization scheme 162742389

awisema:


eanders:
@awisema this looks pretty good.  We'll want one additional bit.

There should be a class method on `Client` to call that will add the `tie_breaker` value to anyone who doesn't have it, ideally without inflating all client objects.  (There's always the possibility an `after_create` won't be called, say if the client records are added in a batch.)  Then we'll need a scheduled task in `schedule.rb` that runs every hour and calls the new method.

app/models/client.rb
```diff
@@ -161,6 +165,20 @@ def self.prioritized match_route:
     end
   end
 
+  # A random number for prioritizations that require a tie-breaker
+  def assign_tie_breaker
+    tie_breaker = rand
+  end
+
+  def self.add_missing_tie_breakers
```
eanders: Nice!

PR #149, opened 2018-12-31
Create homeless set asides route

awisema:
WIP review so we can discuss routes in more detail.


app/models/match_decisions/homeless_set_aside/hsa_accepts_client.rb
```diff
@@ -0,0 +1,110 @@
+module MatchDecisions::HomelessSetAside
```
eanders: This file has changed for the Provider Only Route and needs a quick update.

app/models/match_decisions/homeless_set_aside/hsa_acknowledges_receipt.rb
```diff
@@ -0,0 +1,119 @@
+module MatchDecisions::HomelessSetAside
```
eanders: This file has changed for the Provider Only Route and needs a quick update.

app/mailers/notifications_mailer.rb
```diff
@@ -225,6 +225,15 @@ def hsa_decision_hsp notification
 
   # End Provider Only
 
+  # Set Aside
```
eanders: This file has changed for the Provider Only Route and needs a quick update.

app/models/client_opportunity_match.rb
```diff
@@ -391,6 +391,13 @@ def activate!
     end
   end
 
+  def matched!
+    self.class.trasaction do
+      add_default_contacts!
+      self.send(match_route.initial_decision).initialize_decision!
```
eanders: This is going to "start" the match, putting the first decision into the `:pending` state.  I don't think we want that to happen until the match is activated.

app/models/match_decisions/homeless_set_aside/confirm_hsa_accepts_client_decline_dnd_staff.rb
```diff
@@ -0,0 +1,97 @@
...
+       }
+    end
+
+    def label
+      label_for_status status
+    end
+
+    def label_for_status status
+      case status.to_sym
+      when :pending then "#{_('DND')} to confirm match success"
```
eanders: This isn't correct.  should be `"#{_('DND')} to confirm #{_('Housing Subsidy Administrator')} decline`

app/models/match_decisions/homeless_set_aside/confirm_hsa_accepts_client_decline_dnd_staff.rb
```diff
@@ -0,0 +1,97 @@
...
+
+      def decline_overridden_returned
+        # Re-initialize the previous decision
+        match.hsa_accepts_client_decision.initialize_decision!
+        @decision.uninitialize_decision!(send_notifications: false)
+      end
+
+      def decline_confirmed
+        match.rejected!
+        # TODO maybe rerun the matching engine for that vacancy and client
```
eanders: I don't think this is necessary, nor will it help since the message has been sent.

app/models/match_decisions/homeless_set_aside/hsa_acknowledges_receipt.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    def accessible_by? contact
+      contact.user_can_act_on_behalf_of_match_contacts? ||
+      contact.in?(match.housing_subsidy_admin_contacts)
+    end
+
+    def notifications_for_this_step
+      @notifications_for_this_step ||= [].tap do |m|
+        m << Notifications::HomelessSetAside::MatchInitiationForManualActivation
```
eanders: Pretty sure `MatchInitiationForManualActivation` isn't defined anywhere.  I think this should remain `MatchInitiationForHsa` as the manual activation notification is one step before this.

eanders: Looks like maybe this was supposed to be attached to `MatchInitiationForManualNotification`

app/views/notifications_mailer/match_initiation_for_manual_activation.haml
```diff
@@ -0,0 +1,17 @@
+Hello #{@contact.full_name},
```
eanders: @awisema let's review this in context together, I think the wording needs be updated for the set-asides.

app/models/match_decisions/homeless_set_aside/confirm_hsa_accepts_client_decline_dnd_staff.rb
awisema: Fixed both here and on the provider only route.

PR #150, opened 2019-01-05
Story/add tests for new rules 162895781

awisema:


spec/models/rules/active_in_cohort_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
+
+RSpec.describe Rules::ActiveInCohort, type: :model do
+  before { skip "Requires warehouse classes" }
```
eanders: Since these check against the local to CAS `active_cohort_ids` do we really need to skip all of them?

spec/factories/rules.rb
```diff
@@ -95,4 +111,197 @@
     name 'Minimum occupancy'
     verb 'have'
   end
+  factory :interested_in_cambridge, class: 'Rules::InterestedInNeighborhood' do
```
eanders: Shouldn't this be `interested_in_neighborhood`? The requirement would need a "variable" to specify the specific neighborhood.

spec/models/rules/active_in_cohort_spec.rb
eanders: I see, it's comparing against the ids, which probably wants a DB backing.... maybe `let(:cohort_a)..

spec/models/rules/interested_in_neighborhood_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.describe Rules::InterestedInNeighborhood, type: :model do
+  describe 'clients_that_fit' do
+
+    let!(:cambridge) { create :neighborhood_cambridge }
+    let!(:beacon_hill) { create :neighborhood_beacon_hill }
+
+    let!(:rule) { create :interested_in_cambridge }
```
eanders: This also should be `interested_in_neighborhood`

spec/models/rules/active_in_cohort_spec.rb
awisema: Replaced reified warehouse objects with simple ids to avoid dependency.

PR #151, opened 2019-01-08
Story/ui to handle multiple active matches 162960715

awisema:


app/controllers/opportunity_matches_controller.rb
```diff
@@ -8,14 +8,9 @@ class OpportunityMatchesController < ApplicationController
...
-    @sub_program = @opportunity.sub_program
-    @program = @sub_program.program
-  end
-
-  def all
-    clients_for_route = Client.all
-    @matches = @opportunity.matching_clients(clients_for_route).page(params[:page]).per(25)
+    @actives = @opportunity.active_matches.map { |match| match.client }
+    @availables = @opportunity.matching_clients(clients_for_route)
+    @matches = Kaminari.paginate_array(@actives + @availables).page(params[:page]).per(25)
```
eanders: I don't see a reason to paginate this if we have to load them all anyway.  We should test what happens when 200 clients match though.  Might make the page very slow.

app/controllers/opportunity_matches_controller.rb
```diff
@@ -26,29 +21,68 @@ def closed
     @program = @sub_program.program
   end
 
+  def create
+    client_ids_to_activate = params[:checkboxes].reject { | key, value | value != "1" }.keys.map(&:to_i)
+    client_ids_to_activate.each do | client_id |
+      matches = ClientOpportunityMatch.where(client_id: client_id, opportunity_id: @opportunity, closed: false)
```
eanders: I believe you could use `first_or_create` here.  It will be much cleaner.

app/models/opportunity.rb
```diff
@@ -14,6 +14,8 @@ class Opportunity < ActiveRecord::Base
   has_one :sub_program, through: :voucher
 
   has_one :active_match, -> {where(active: true, closed: false)}, class_name: 'ClientOpportunityMatch'
```
eanders: We should comment this out (deprecate it) since it shouldn't be valid anymore.  That should help find the places where we use it too.

app/controllers/opportunity_matches_controller.rb
awisema: I experimented with using a back port of Rails 5 or to combine the actives and available, but a limitation of that functionality is the two sets must differ only by the where clauses...

Also experimented with plucking the ids for the two sets and then reifying the objects based on those, but couldn't figure out how to order the results... Currently, the active matches appear at the top, but maybe that isn't required, the table could just be ranked by the route, and highlight the active matches?

app/models/opportunity.rb
awisema: Commented out and made an initial pass through the code to find dependencies, replacing them with the first item in `active_matches` - the pages will need to be inspected to see if they need to be changed to display multiple active matches on an opportunity.

PR #155, opened 2019-01-31
Story/sort opportunity matches so nulls fall to bottom 163592739

awisema:


app/models/match_prioritization/days_homeless_last_three_years_random_tie_breaker.rb
```diff
@@ -5,8 +5,9 @@ def self.title
       'Days homeless in the last three years with random tie-breaker'
     end
 
-    def self.slug
-      'homeless-days-last-three-years-random-tie-breaker'
+    def self.prioritization_for_clients(scope)
+      scope.order(c_t[:days_homeless_in_last_three_years].desc.to_sql + ' NULLS LAST').
```
eanders: I think this needs an Arel sql literal incantation.  (or will for rails 5, and we might as well add it now.)

PR #167, opened 2019-03-12
Offline changes

eanders:
This completes:
https://www.pivotaltracker.com/story/show/163399682
https://www.pivotaltracker.com/story/show/163150649
https://www.pivotaltracker.com/story/show/162914245
https://www.pivotaltracker.com/story/show/162739266


app/models/deidentified_client.rb
```diff
@@ -7,4 +7,77 @@ class DeidentifiedClient < NonHmisClient
     text.strip!
     where(search_alternate_name(text))
   end
+
+  def download_headers
```
awisema: I think we have used a hash to combine this and `download_data` elsewhere?

app/models/program.rb
```diff
@@ -131,6 +131,17 @@ def self.associations_adding_services
     [:funding_source]
   end
 
+  def self.sort_options
+    [
+      {title: 'Program A-Z', column: 'program_id', direction: 'asc', order: 'LOWER(programs.name) ASC', visible: true},
+      {title: 'Program Z-A', column: 'program_id', direction: 'desc', order: 'LOWER(programs.name) DESC', visible: true},
+      {title: 'Sub-Program A-Z', column: 'sub_program_id', direction: 'asc', order: 'LOWER(sub_programs.name) ASC', visible: true},
+      {title: 'Sub-Program Z-A', column: 'sub_program_id', direction: 'desc', order: 'LOWER(sub_programs.name) DESC', visible: true},
+      {title: 'Building A-Z', column: 'bilding_id', direction: 'asc', order: 'LOWER(buildings.name) ASC', visible: true},
```
awisema: building_id?

app/models/deidentified_client.rb
eanders: We have (at least in the warehouse.)   At this time this is only for the export, so it seemed somewhat less necessary to build all of the hashes.  That said, it would make it less error prone.

app/views/clients/index.html.haml
```diff
@@ -14,15 +14,10 @@
...
-    %a.nav-link.active#available-clients{href: '#available', data: {toggle: :tab}, role: :tab, aria: {controls: :available, selected: true}}
-      Available Clients
-  %li.nav-item
-    %a.nav-link#unavailable-clients{href: '#unavailable', data: {toggle: :tab}, role: :tab, aria: {controls: :unavailable, selected: false}}
-      Unavailable Clients
+  %li.nav-item= link_to 'Available Clients', clients_path, class: 'nav-link active'
+    
+  %li.nav-item= link_to 'Unavailable Clients', unavailable_clients_path, class: 'nav-link'
 
 .tab-content
```
awisema: Is this required having broken unavailable clients out onto another page?

app/models/program.rb
eanders: yep.  fixed

app/views/identified_clients/index.xlsx.axlsx
```diff
@@ -0,0 +1,75 @@
+wb = xlsx_package.workbook
+wb.add_worksheet(name: client_type) do |sheet|
+  title = sheet.styles.add_style(sz: 12, b: true, alignment: {horizontal: :center})
+  sheet.add_row(
+    [
```
awisema: Use `download_headers` (or the keys if made into a map)?

eanders: Ah, this file should have been removed.

PR #174, opened 2019-04-09
Expose match cancellation as a result of parking

awisema:


app/models/client.rb
```diff
@@ -306,12 +306,16 @@ def make_unavailable_in_all_routes
...
-  def unavailable(permanent:false)
+  def unavailable(permanent:false, user:nil)
     active_match = client_opportunity_matches.active.first
     Client.transaction do
       if active_match.present?
         opportunity = active_match.opportunity
         active_match.canceled!
+        MatchEvents::Parked.create!(
+            match_id: active_match.id,
+            contact_id: user&.id
```
eanders: I believe this should be `user&.contact&.id`

app/models/client.rb
```diff
@@ -382,8 +386,20 @@ def self.possible_availability_states
     return states
   end
 
+  def parked?
+    prevent_matching_until.present? && prevent_matching_until > Date.today
+  end
+
+  def available?
```
eanders: This will want to have a different name, maybe `available_for_matching?`.  There's a column on client `available` which magically responds to `available?`

PR #176, opened 2019-04-12
Stories/automatically add tags 165252256

eanders:


app/views/tags/_form.haml
```diff
@@ -4,6 +4,7 @@
       = f.error_notification
       .form-inputs
         = f.input :name
+        = f.input :rrh_assessment_trigger, label: _('should this tag be added to anyone with an assessment score?')
```
awisema: Capitalize 'Should'?

PR #181, opened 2019-04-24
Only show assigned contacts on set-asides match

awisema:


app/views/match_contacts/_match_contacts_list.haml
```diff
@@ -1,5 +1,10 @@
 %ul.jMatchContactList.c-detail-group__content.list-unstyled
-  - match.contact_titles.each do |contact_type, title|
+  - if match.match_route.show_default_contact_types
```
eanders: Can we delegate the contact_titles back to the route?  Seems like a useful place to move complexity out of the view.

PR #183, opened 2019-04-26
Building and unit pickers on client housed step

awisema:


eanders:
When accessing the match using the links (not logging in) the unit drop-down is not updating.
Screen Shot 2019-05-01 at 8 34 40 AM

eanders:
When logged in as an administrator, the size of the drop-downs is too small.  It looks like they are set to 50% width, which works when there is only one column, but admins get 2.
Screen Shot 2019-05-01 at 8 36 09 AM

eanders:
If the unit is changed, we should add an event to the history stating that it was changed.
Screen Shot 2019-05-01 at 8 38 07 AM

PR #187, opened 2019-05-06
Fix removing and re-adding a variable rule

awisema:


eanders:
Nice, that solution was much cleaner than I was expecting.

PR #198, opened 2019-05-18
story-166032579

andrewbraunstein:
Added Adit buttons when clients are non-HMIS cliewnts.  however, there are currently no breadcrumbs that exist with this functionality.

### DIFF
--- a/app/models/non_hmis_client.rb
+++ b/app/models/non_hmis_client.rb
@@ -148,9 +148,9 @@ class NonHmisClient < ActiveRecord::Base
       delete_all
 
     # update or add for all NonHmisClients
-    NonHmisClient.all.each do |client|
+    NonHmisClient.all.find_each do |client|
       project_client = ProjectClient.where(
-        data_source_id: DataSource.non_hmis.select(:id),
+        data_source_id: DataSource.non_hmis.pluck(:id),
         id_in_data_source: client.id
       ).first_or_initialize
       client.populate_project_client(project_client)
--- a/app/models/non_hmis_client.rb
+++ b/app/models/non_hmis_client.rb
@@ -153,7 +153,7 @@ class NonHmisClient < ActiveRecord::Base
     # remove unused ProjectClients
     ProjectClient.where(
         data_source_id: data_source_id).
-        where.not(id_in_data_source: NonHmisClient.deidentified_data_source.select(:id)).
+        where.not(id_in_data_source: NonHmisClient.select(:id)).
         delete_all
 
     # update or add for all NonHmisClients
--- a/app/models/non_hmis_client.rb
+++ b/app/models/non_hmis_client.rb
@@ -1,7 +1,7 @@
 class NonHmisClient < ActiveRecord::Base
   has_one :project_client, ->  do
     where(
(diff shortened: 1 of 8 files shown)


app/models/non_hmis_client.rb
```diff
@@ -1,7 +1,7 @@
 class NonHmisClient < ActiveRecord::Base
   has_one :project_client, ->  do
     where(
-        data_source_id: DataSource.where(db_identifier: 'Deidentified').select(:id),
+        data_source_id: NonHmisClient.data_source_id
```
eanders: This change forces an extra SQL query that is unnecessary.

app/models/non_hmis_client.rb
```diff
@@ -73,6 +73,10 @@ def cohort_names
     arel_table[:client_identifier].lower.matches("%#{name.downcase}%")
   end
 
+  def self.data_source_id
```
eanders: Perhaps this could be 
```
def self.deidentified_data_source
  DataSource.where(db_identifier: 'Deidentified')
end
```
Then above and below you could call `where(data_source_id: NonHmisClient.deidentified_data_source.select(:id))`
This will centralize the complexity, but keep a relation and prevent an extra query.

app/views/clients/_client_list.haml
```diff
@@ -68,9 +70,19 @@
...
                   = masked_ssn client.ssn
                   = client_data_quality client, :ssn
-              %th.text-center
+              %td.text-center
                 = checkmark client.veteran
-              %th.text-center
+              %td.text-center
                 = checkmark client.chronic_homeless
+              %td
+                - if client.project_client.data_source_id == non_hmis_source_id
```
eanders: This should be moved to method on project client `source_non_hmis?` (or something similar).  Both this and `is_deidentified?` can be added to client as well and delegated to the project_client.

app/views/clients/_client_list.haml
```diff
@@ -68,9 +70,19 @@
...
+                  - if client.project_client.is_deidentified?
+                    = link_to edit_deidentified_client_path(client.project_client.id_in_data_source), class: ['btn', 'btn-sm', 'btn-secondary'] do
```
eanders: The logic for where the link goes can be moved up into the `NonHmisClient` models in `#edit_path`.  Each should return arrays that would work with `polymorphic_path` so we can move the complexity out of the view.

eanders:
@andrewbraunstein please also check and fix the Travis failures above.

app/models/non_hmis_client.rb
andrewbraunstein: Putting it back to the way it was.

app/views/clients/_client_list.haml
andrewbraunstein: I made the change as requested.  I had originally done it this way so that the non_hmis_source_id would only need to be evaluated once and not in every iteration of the loop.

app/views/clients/_client_list.haml
andrewbraunstein: I was able to leverage existing polymorphism in the NonHmis controller to achieve the same effect.

PR #204, opened 2019-05-23
Importer for set asides

awisema:


awisema:
Test failures are addressed in the upstream branch where they came from, and are not merged here to keep the review small.

awisema:
Test fails were addressed in the source branch, and have not been merged here to keep the diffs small.

eanders:
One more thing. Let's add a few tests to verify that the imported file creates the expected number of records and that if there are existing records they match appropriately.

spec/models/imported_clients_csv_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
```
eanders: Can we add ~5 rows to the sample files so we don't have an N of 1 (those often have their own sets of issues.) 

Also, let's add a test where it matches an existing ImportedClient so we can make sure that's working. 

Finally, a test that verifies that the people we expect are the people we actually get (compare emails maybe?), especially when one already exists.

PR #227, opened 2019-08-12
Stalled match config per route

awisema:


app/controllers/admin/match_routes_controller.rb
```diff
@@ -22,6 +22,16 @@ def update
       respond_with(@route, location: admin_match_routes_path)
     end
 
+    def stall_intervals
```
eanders: Since this isn't used in the controller, let's move it to the model.  Then we can reference is there if necessary (say for validations) and can call it in the view with `@route.stall_invervals`.

PR #237, opened 2019-09-09
Code to support Decline Report in the warehouse

eanders:
This needs to be run after merging and deploying `stories/decline-reasons-report-168038272` in the warehouse code base.


app/models/warehouse/build_report.rb
```diff
@@ -138,6 +139,13 @@ def fill_cas_report_table!
                 end
               end
 
+              # If this a canceled or declined decision, log who did it
+              event_contact = if decision.status.in?(['declined', 'canceled'])
+                decision.decision_action_events.select do |da|
```
awisema: Use `detect` instead of `select` so you don't need the `first`?

eanders: Of course.  Thanks

PR #248, opened 2019-09-27
Allow HSA to cancel other opportunity related matches

awisema:


app/models/client_opportunity_match.rb
```diff
@@ -554,6 +554,18 @@ def succeeded!
     end
   end
 
+  def cancel_opportunity_related_matches
```
eanders: add a bang!
`def cancel_opportunity_related_matches!`

PR #252, opened 2019-10-02
Revert "Revert "Make changing route on existing program with matches not crash""

eanders:
Reverts greenriver/boston-cas#251

Revert the revert


eanders:
Pending tests, this looks good.  Thanks @awisema

PR #254, opened 2019-10-02
Vacancies page update

awisema:


app/views/opportunities/index.haml
```diff
@@ -18,73 +18,63 @@
...
+              Sub-Program
             %th
               Program
             %th
-              Sub-Program
+              Status
             %th
-              Voucher Type
-            %th= "#{Building.model_name.human} & Unit"
+              Created on
```
eanders: Change Matches header to "Prioritized Clients", and move to the far right.

PR #276, opened 2019-11-22
Unavailable for expiration

eanders:
This completes this story: https://www.pivotaltracker.com/story/show/169844453.

It adds an optional expiration length for "unavailable as candidate fors" and provides a means for someone to manually make a client available on a route.

### DIFF
--- a/app/controllers/unavailable_on_route_controller.rb
+++ b/app/controllers/unavailable_on_route_controller.rb
@@ -15,7 +15,6 @@ class UnavailableOnRouteController < ApplicationController
       flash[:notice] = "#{@client.name} is now available to match on the #{@unavailable_for.route&.title}"
     else
       flash[:error] = "Unable to make #{@client.name} available on the #{@unavailable_for.route&.title}"
-      render :new
     end
     redirect_to client_path(@client)
   end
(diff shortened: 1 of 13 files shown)


app/models/unavailable_as_candidate_for.rb
```diff
@@ -10,6 +10,7 @@
 class UnavailableAsCandidateFor < ActiveRecord::Base
   belongs_to :client
   belongs_to :route, class_name: 'MatchRoutes::Base', primary_key: :type, foreign_key: :match_route_type
+  has_paper_trail
```
awisema: Is this just to capture the whodunnit? Doesn't seem like there is going to be any history except creation and deletion events...

eanders: Yes, this is just so we can say when a client was unavailable (since that can now be removed).  And to capture who made them available when they match something they shouldn't have.

app/controllers/unavailable_on_route_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  before_action :some_clients_editable!
+  before_action :set_client
+  before_action :set_unavailable_for
+
+  def destroy
+    if @unavailable_for.destroy
+      flash[:notice] = "#{@client.name} is now available to match on the #{@unavailable_for.route&.title}"
+    else
+      flash[:error] = "Unable to make #{@client.name} available on the #{@unavailable_for.route&.title}"
+      render :new
```
awisema: Where will this end up?

eanders: No where!   Good catch, that was a bad copy paste job.

PR #278, opened 2019-11-25
Story/stop client match and requeue 169739605

awisema:


app/views/matches/_current_step_info.haml
```diff
@@ -15,13 +15,34 @@
             = @match.current_decision.try(:actor_type) || 'N/A'
       = render 'match_list_base/current_step', match: @match
   - else
-    %p This match is not active
+    -  description = 'This match is not active'
+    - if can_reopen_matches? && ! @match.can_be_reopened?
+      - description += ', and cannot be re-opened because'
```
eanders: Can we move the calculation of what `description` is into the model?  It becomes much more complicated than it needs to be in the view.  Also, if you treat description as an array and just join it with `, and` at the you won't need to keep track of `emit_and`

app/models/client_opportunity_match.rb
```diff
@@ -453,6 +453,36 @@ def expire_all_notifications(on: 1.week.from_now.to_date)
...
+  def can_be_reopened?
+    return false if would_be_client_multiple_match
+    return false if would_be_opportunity_multiple_match
+
+    true
+  end
+
+  def reopen!
+    client.make_unavailable_in(match_route: match_route)
+    update(closed: false, active: true, closed_reason: nil)
```
eanders: Let's log an event here indicating the match was re-opened.

app/models/client_opportunity_match.rb
```diff
@@ -453,6 +453,36 @@ def expire_all_notifications(on: 1.week.from_now.to_date)
...
+    current_decision.update(status: :pending)
```
eanders: I *think* if the current decision was DND accepts match there's no mechanism to "backup."  So, this may need to move the match back 2 steps in some situations.

awisema: I added the back button to this step.

awisema:
This should be ready for re-review.

PR #281, opened 2019-12-04
Story/not all closed matches get red exes

awisema:


eanders:
Thanks!  Looks great.

PR #293, opened 2020-01-07
Story/pathways assessment for non hmis clients 169375966

eanders:


app/controllers/identified_clients_controller.rb
```diff
@@ -67,42 +61,53 @@ def client_source
...
+          :domestic_violence,
+          :interested_in_set_asides,
+          :youth_rrh_aggregate,
+          :dv_rrh_aggregate,
+          :veteran_rrh_desired,
+          :rrh_th_desired,
+          :sro_ok,
+          :other_accessibility,
+          :disabled_housing,
+          neighborhood_interests: [],
```
eanders: This looks very much the same as the deidentified version.  We should probably move it to the model to DRY it up a bit.

app/controllers/non_hmis_clients_controller.rb
```diff
@@ -94,9 +100,14 @@ def sorter
     return sort_string
   end
 
+  def build_assessment
+    assessment_type.constantize.new
```
eanders: where does `assessment_type` come from?  It will need to be from a whitelist if we're going to call `constantize`.

eanders: Found it.

app/controllers/identified_clients_controller.rb
```diff
@@ -23,20 +23,14 @@ def destroy
     respond_with(@non_hmis_client, location: identified_clients_path)
   end
 
+  def assessment_type
+    Config.get(:identified_client_assessment) || 'IdentifiedClientAssessment'
```
eanders: This will need to hit a whitelist.  Probably something on `NonHmisAssessment` like.
`def self.assessment_type(type)
    {
      'NonHmisAssessment' => NonHmisAssessment,
      'IdentifiedClientAssessment' => IdentifiedClientAssessment,
      'IdentifiedPathwaysAssessment' => IdentifiedPathwaysAssessment,
    }.freeze[type] || NonHmisAssessment
  end`

app/models/deidentified_client.rb
```diff
@@ -62,37 +64,37 @@ def download_data
...
+      current_assessment.date_days_homeless_verified,
+      current_assessment.who_verified_days_homeless,
+      current_assessment.income_total_monthly,
+      current_assessment.disabling_condition,
+      current_assessment.physical_disability,
+      current_assessment.developmental_disability,
+      current_assessment.domestic_violence,
       created_at,
-      updated_at,
+      current_assessment.updated_at,
```
eanders: Let's standardize this in `NonHmisClient`  I believe the differences between identified and non-identified are small enough we could just just remove those that don't apply or, just leave them blank when they don't apply.  Also, it looks like de-identified gets VI-SPDAT, but those should probably be determined by the site config.

app/models/identified_pathways_assessment.rb
```diff
@@ -0,0 +1,28 @@
...
+      self.youth_rrh_aggregate = 'adult'
+    end
+
+    if dv_rrh_desired? && rrh_desired?
+      self.dv_rrh_aggregate = 'both'
+    elsif dv_rrh_desired?
+      self.dv_rrh_aggregate = 'dv'
+    elsif rrh_desired?
+      self.dv_rrh_aggregate = 'non-dv'
+    end
```
eanders: Is there a difference here between identified and de-identified?  If not, let's move it to `IdentifiedClientAssessment`

app/models/imported_client.rb
```diff
@@ -4,17 +4,12 @@ class ImportedClient < NonHmisClient
       data_source_id: DataSource.where(db_identifier: ImportedClient.ds_identifier).select(:id),
     )
   end, foreign_key: :id_in_data_source, required: false
+  has_many :client_assessments, class_name: 'ImportedClientAssessment', foreign_key: :non_hmis_client_id, dependent: :destroy
+  has_one :current_assessment, -> { order(created_at: :desc) }, class_name: 'ImportedClientAssessment', foreign_key: :non_hmis_client_id
```
eanders: I _think_ we've determined ordering by id is generally faster and returns the same result.

app/models/imported_clients_csv.rb
```diff
@@ -61,13 +61,14 @@ def import(agency)
...
+        assessment = client.current_assessment || client.client_assessments.build
+
         timestamp = convert_to_time(row[FORM_TIMESTAMP])
-        if client.imported_timestamp.nil? || timestamp > client.imported_timestamp
+        if assessment.imported_timestamp.nil? || timestamp > assessment.imported_timestamp
           @clients << client
-          @touched += 1 if client.imported_timestamp.present?
-          client.update(
+          @touched += 1 if assessment.imported_timestamp.present?
+          assessment.update_attributes(
```
eanders: `update` and `update_attributes` are equivalent.  Did you mean `update_columns`? https://davidverhasselt.com/set-attributes-in-activerecord/

app/models/non_hmis_client.rb
```diff
@@ -10,6 +10,14 @@ class NonHmisClient < ActiveRecord::Base
   end, foreign_key: :id_in_data_source, required: false
   has_one :client, through: :project_client, required: false
   has_many :client_opportunity_matches, through: :client
+
+  has_many :non_hmis_assessments
+
+  def current_assessment
```
eanders: Maybe this should be `has_one` instead of `def`?

app/models/deidentified_client.rb
eanders: See `Config.get(:non_hmis_fields).include?('vispdat_score')` in `app/views/deidentified_clients/_deidentified_client_assessment.haml` for examples.

app/views/identified_clients/_assessments.haml
```diff
@@ -0,0 +1,7 @@
+- if @non_hmis_client.non_hmis_assessments.count > 1
```
eanders: If we set `@assessments = @non_hmis_client.non_hmis_assessments`, `@client = @non_hmis_client`, and `@path = identified_client_path` in the controller could we streamline the views and just have one instance of most of them?

app/views/imported_clients/_table.haml
```diff
@@ -21,7 +21,7 @@
                 = link_to({action: :edit, id: imported_client}) do
                   = imported_client.first_name
               %td= checkmark imported_client.family_member
-              %td= imported_client.days_homeless_in_the_last_three_years
+              %td= imported_client.current_assessment.days_homeless_in_the_last_three_years
```
eanders: Is it worth setting up delegation for all of these things that used to exist on the imported client.  Delegate to the current_assessment.

db/migrate/20191206143217_relax_non_null_constraint_on_neighborhood_interests.rb
```diff
@@ -0,0 +1,8 @@
+class RelaxNonNullConstraintOnNeighborhoodInterests < ActiveRecord::Migration
+  def up
+    change_column :non_hmis_clients, :neighborhood_interests, :json, null: true
```
eanders: This should probably be `:jsonb`.

db/migrate/20200117184252_set_active_and_eligible_from_available_for_non_hmis_clients.rb
```diff
@@ -0,0 +1,6 @@
+class SetActiveAndEligibleFromAvailableForNonHmisClients < ActiveRecord::Migration
+  def up
+    NonHmisClient.update_all('active_client=available')
```
eanders: why not? `NonHmisClient.update_all(active_client: :available)`

PR #295, opened 2020-01-09
Calculate and save off match census data once a day

eanders:


app/models/match_census.rb
```diff
@@ -0,0 +1,55 @@
+###
+# Copyright 2016 - 2019 Green River Data Analysis, LLC
```
awisema: 2020 :-)

app/models/match_census.rb
```diff
@@ -0,0 +1,55 @@
...
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/master/LICENSE.md
+###
+
+class MatchCensus < ActiveRecord::Base
+  belongs_to :opportunity
+  belongs_to :match, class_name: 'ClientOpportunityMatch', required: false
+
+  def self.populate!
+    # Add all opportunities that have not been ma
```
awisema: ?

app/models/match_census.rb
```diff
@@ -0,0 +1,55 @@
...
+      populate_from_opportunity!(opp)
+    end
+  end
+
+  def self.populate_from_opportunity! opp
+    self.transaction do
+      # Clear any existing data for this opportunity on this day, and rebuild
+      where(opportunity_id: opp.id, date: Date.current).delete_all
+      clients_for_route = Client.available_for_matching(opp.match_route)
+      available_client_ids = opp.matching_clients(clients_for_route).pluck(:id)
```
awisema: Add comment noting that this is sorted by the prioritization function on the match route?

app/models/match_census.rb
eanders: Yep, that's my next PR

PR #306, opened 2020-02-06
170856050 pathways updates  controller

eanders:


eanders:
@jgoley, I believe all work moved to https://github.com/greenriver/boston-cas/pull/309, but it looks like there's some work here.  Can you confirm, this is incomplete or unnecessary now?

jgoley:
@eanders I think the work started in this is probably a good direction to head but not crucial. It starts to refactor out Assessments from the Client controllers. It may be more work than necessary though. The view structure is roughly in place but there's still work to be done in the controllers.

eanders:
Thanks @jgoley!

eanders:
Closing, most of this change has been incorporated, or deemed no longer necessary.

PR #307, opened 2020-02-06
Select2 as simple form input

jgoley:
@eanders This PR creates a select2 simple_form input type by extending `CollectionSelectInput` it instantiates a JS wrapper class that receives the select's ID and initializes select2.

I left select2 being imported within application.js since it looks like there are instances throughout.
I started taking a look at refactoring all the old instances and can go ahead, but that's a pretty major change—should we take an incremental approach?

I also setup a styleguide page for a form: `/style_guide/form` with an example of select2 within a `simple_form`


eanders:
This looks great.  Thanks @jgoley.  I agree, let's make it an incremental change.  
I'll change a page of them now and we can slowly migration to `as: :select2` going forward.

PR #310, opened 2020-02-20
Use alternate rule names on units

awisema:
Should we use the alternate names anywhere else?


eanders:
@awisema I believe this needs to be updated here as well app/views/matches/_rules.haml.  I believe this will fix both the prioritized clients view and the match itself, which still uses the name instead of the alternate name....  Might be harder because you'll have to determine where the rule is coming from.

app/views/requirement_manager/_form_fields.haml
```diff
@@ -1,5 +1,6 @@
 - help_text ||= ''
 - hide_inherited ||= false
+- on_unit ||= false
```
eanders: This is fine, but only because the default is false.  We need to be careful with ||= and booleans.  A false passed in will evaluate to false.  If the default were true, then you'd never be able to pass in a false.  Might be safer to write `on_unit = false unless defined?(on_unit)`

PR #311, opened 2020-02-20
Default program contacts on sub programs

awisema:


app/models/contact.rb
```diff
@@ -19,6 +19,7 @@ class Contact < ApplicationRecord
   has_many :client_contacts, inverse_of: :contact, dependent: :destroy
   has_many :opportunity_contacts, inverse_of: :contact, dependent: :destroy
   has_many :program_contacts, inverse_of: :contact, dependent: :destroy
```
eanders: This should probably be removed.

eanders:
@awisema, it looks like there may be a fair amount of program_contact code still in there.  Can we tear that out?

PR #317, opened 2020-02-25
Filter vacancies page

awisema:


app/controllers/opportunities_controller.rb
```diff
@@ -5,6 +5,7 @@
 ###
 
 class OpportunitiesController < ApplicationController
+  include ArelHelper
```
eanders: Let's put this directly in ApplicationController, and ApplicationRecord

app/views/opportunities/_tabs.haml
```diff
@@ -0,0 +1,80 @@
...
+                    - elsif opportunity.available || opportunity.voucher.available
+                      Available
+                  %td{:class => deleted}= opportunity.voucher&.available_at&.to_date
+                  %td.text-center
+                    - if opportunity.show_alternate_clients_to?(current_user)
+                      = link_to opportunity_matches_path(opportunity) do
+                        .badge.badge-success= opportunity.matching_clients.count
+        %p= paginate @opportunity_tabs[dashed_route_name]
+
+= content_for :page_js do
```
eanders: This isn't correctly remembering tabs.  It's putting a hash at the end of the params.  It should go immediately after the path.  Better yet, we should pass the route as a param so that we don't need any of the JS and push state.

PR #321, opened 2020-02-27
Replace route filter with tabs

awisema:


eanders:
This looks great.  Thanks @awisema!

PR #327, opened 2020-03-10
DV RRH Rule and supporting fields

eanders:


db/migrate/20200310001320_add_dvrrh_fields.rb
```diff
@@ -0,0 +1,16 @@
+class AddDvrrhFields < ActiveRecord::Migration[6.0]
+  def change
+    # remove some unused pathways columns
+    remove_column :project_clients, :pathways_domestic_violence, :boolean, defaul: false, null: false
+    remove_column :clients, :pathways_domestic_violence, :boolean, defaul: false, null: false
```
awisema: Shouldn't matter -- but there is a typo in `default`

eanders: Thanks

PR #329, opened 2020-03-11
Add drill down to CAS dashboard overview

awisema:


app/controllers/reports/dashboards/overviews_controller.rb
```diff
@@ -5,14 +5,68 @@
...
+
   before_action :require_can_view_reports!
+  before_action :set_report
 
   def index
+  end
+
+  def details
+    @section = sections.detect { |name| params[:section].to_sym == name }
+    @key = params[:key]
```
eanders: Let's be extra careful and limit these to `data.keys.detect...` and whatever the equivalent for sub-keys would bey, maybe `data.values.keys?`  Then we can return @clients = nil or some such if we don't have valid keys.

app/views/reports/dashboards/overviews/details.haml
```diff
@@ -0,0 +1,16 @@
...
+
+%table.table
+  %thead
+    - details_columns.each do |name|
+      %th= name.to_s.humanize
+  %tbody
+    - @clients.each do |client|
+      %tr
+        - details_columns.each do |name|
+          %td= client.send(name)
```
eanders: Let's call `public_send` here.  Probably doesn't matter, but it will avoid accidentally calling private methods.

app/views/reports/dashboards/overviews/details.haml
```diff
@@ -0,0 +1,25 @@
...
+= content_for :modal_title, title
+
+%table.table
+  %thead
+    - details_columns.each do |name|
+      %th= name.to_s.humanize
+  %tbody
+    - @clients.each do |client|
+      %tr
+        - if client.accessible_by_user?(current_user)
```
eanders: This is going to be a relatively expensive N+1.  Can we figure out how to select it as a column or in a batch and pre-fetch `Client.where(id: client_ids).visible_by(user).pluck(:id)` and just do a lookup?

awisema: Fixed.

PR #330, opened 2020-03-12
Transition from parking lot to unavailable fors

eanders:


app/controllers/reports/parked_clients_controller.rb
```diff
@@ -7,20 +7,43 @@
...
-        clients = parked_scope.non_confidential.full_release
-      end
+      @active_tab = params[:tab] || route_to_html_id(@routes.first)
+      clients = parked_scope
+      clients = clients.non_confidential.full_release unless can_view_all_clients?
+
       @clients = clients.
-        order(prevent_matching_until: :asc).
+        joins(:unavailable_as_candidate_fors).
+        order('unavailable_as_candidate_fors.expires_at asc').
```
awisema: Add `uacf_t` to `ArelHelper` and replace the string with `uacf_t[:expires_at].asc`?

app/models/client.rb
```diff
@@ -501,11 +500,7 @@ def available_text
         'Fully matched'
       end
     else
-      if parked?
-        "Parked until #{prevent_matching_until}"
-      else
-        'Not available'
-      end
+      'Not available'
```
awisema: Maybe bring back the if and return`Not available on any route`? It seems like there is a difference between these two cases that might matter to users.

db/migrate/20200312144826_update_existing_unavailable_fors.rb
```diff
@@ -0,0 +1,14 @@
+class UpdateExistingUnavailableFors < ActiveRecord::Migration[6.0]
+  def up
+    expiration_length = Config.get(:unavailable_for_length)
+    return if expiration_length.zero?
+
+    UnavailableAsCandidateFor.find_each do |u|
+      u.update(expires_at: u.created_at + expiration_length.days)
+    end
+
+    Client.where("prevent_matching_until > ?", Date.current).find_each do |client|
```
awisema: Seems to me that we should either move all parked client, not just the future ones... If there are expired parked clients, should we be cleaning them up?

eanders: Those should have been cleaned up already.  There's a nightly process that takes care of that currently.  Also, we shouldn't be looking at the field anymore, and it should probably be removed eventually.

app/models/client.rb
eanders: There is a difference, but we can no longer include a date because each "parking" can include its own date.

awisema: Yes, I just wondered if we should return different language so the user might think to look at the routes.. I'm tempted to suggest something like "Not available on any route until {earliest_date}" but that might be misleading...

db/migrate/20200312144826_update_existing_unavailable_fors.rb
awisema: If they are cleaned up, can we just do `Client.where.not(prevent_matching_until: nil)`?

PR #335, opened 2020-03-19
Stories/rank rule

eanders:


app/models/rules/rank_below.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def display_for_variable value
+    value.to_i
+  end
+
+  def clients_that_fit(scope, requirement, opportunity)
+    return scope unless opportunity
+
+    if Client.column_names.include?(:tags.to_s)
+      # FIXME, how do we determine the route...
```
awisema: I think you did that 😄

spec/models/rules/rank_below_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+RSpec.describe Rules::RankBelow, type: :model do
+  MatchRoutes::Base.ensure_all
+  describe 'clients_that_fit' do
+    let!(:rule) { create :rank_below }
+    let!(:bob) { create :client, first_name: 'Bob', tags: {1 => 5} }
+    let!(:bob_match) { create :successful_client_opportunity_match, client: bob }
+
+    let!(:roy) { create :client, first_name: 'Roy', tags: {1 => 11} }
+    let!(:hsa_decline_reason) { create :hsa_decline_reason, ineligible_in_warehouse: true }
```
awisema: Isn't this just for the Pathways eligibility rule, and unneeded here?

PR #369, opened 2020-09-09
Include new columns from De-identified spreadsheet

eanders:


app/models/non_hmis_client.rb
```diff
@@ -182,7 +182,7 @@ def set_project_client_fields project_client
     project_client.health_prioritized = health_prioritized
     project_client.hiv_positive = hiv_aids
     project_client.is_currently_youth = current_assessment&.is_currently_youth || false
-
+    project_client.older_than_65 = current_assessment&.older_than_65
```
awisema, line 185: Should this have `|| false` like the line above?

spec/models/rules/age_greater_than_sixty_five_spec.rb
```diff
@@ -18,7 +19,7 @@
 
     context 'when positive' do
       it 'matches 2' do
```
awisema: 'matches 3' - or something like 'matches all over 65 clients'...

awisema (changes requested):

app/models/non_hmis_client.rb
eanders: no because we don't want to assume people are under 65.  It handles null, whereas the other rule does not.

PR #371, opened 2020-09-23
Implement help

tquang1708:
Ported help from HMIS Warehouse to CAS. Everything is 99% functional except that when you create/edit a help using the sidebar form the page does not reload after when you hit Submit (the help document will still be created/updated accordingly).


eanders (approved):
Looks great! Thanks.

PR #379, opened 2020-11-03
Story/route 4 notifications 175313043

eanders:


app/models/notifications/four/match_recommendation_housing_subsidy_admin.rb
```diff
@@ -0,0 +1,37 @@
...
+    def event_label
+      "#{_('Housing Subsidy Administrator')} notified of match recommendation"
+    end
+
+    def decision
+      match.four_match_recommendation_shelter_agency_decision
+    end
+
+    def show_client_info?
+      true
```
eanders, line 25: I think this should be false?

PR #393, opened 2021-01-08
Another VI-SPDAT prioritization scheme

eanders:


app/models/project_client.rb
```diff
@@ -89,13 +89,18 @@ def shelter_agency_contacts
     end&.compact || []
   end
 
-  def self.calculate_vispdat_priority_score(vispdat_score:, days_homeless:, veteran_status:)
+  def self.calculate_vispdat_priority_score(vispdat_score:, days_homeless:, veteran_status:, family_status: false, youth_status: false)
```
awisema: Maybe don't include the defaults? Seems like if we fail to pass these values, it could result in hidden bugs.

awisema (approved):

PR #396, opened 2021-01-13
port change on account password controller from hmis to cas

tquang1708:
https://www.pivotaltracker.com/story/show/176485691


tquang1708:
It seems the failed check was because of a vulnerability with nokogiri, which seems to have gone away in the Gem Update branch

eanders (changes requested):
This looks like a good start.  Can you track down the `users/_password_rules.haml` file from the warehouse, which gets included in the form?  It would be great to implement `User.password_rules` to expose what the rules are on the form.

eanders:
@tquang1708 , this change also seems to have broken one of the tests.  Can you investigate if behavior changed and we need to update the test, or if the test is failing because we broke functionality?

tquang1708:
I only have it shows the "pwned password" and the "password length" rule for now since that's currently the only two active on CAS, and commented everything else out.

tquang1708:
Commented out the test for now. Before it was expecting a redirect (300) to the edit page if password field is empty, but now it just renders the edit form with the errors (200). I don't know exactly what's the right thing to do in this case so I'll just leave the comment here.

eanders:
Let's just change the test to confirm a 200 and the error text is present on the page.

PR #398, opened 2021-01-15
Audit log UI

tquang1708:
[https://www.pivotaltracker.com/story/show/176485717]

I have only ported over the Audit Log, not the Edit History or the Location Tracker.


eanders:
Screen Shot 2021-01-19 at 9 01 22 AM
The Audit button should only show if the permission has been granted.
Screen Shot 2021-01-19 at 9 01 12 AM

Types aren't showing up in the log.  Let's remove that column.
Screen Shot 2021-01-19 at 9 01 43 AM

app/views/admin/audits/_user_tabs.haml
```diff
@@ -0,0 +1,7 @@
+%ul.nav.nav-tabs
+  %li.nav-item{role: "presentation", class: ('active' if current_page?(admin_user_audit_path()))}
+    = link_to 'Audit Trail', admin_user_audit_path(@user), class: 'nav-link'
+  -# %li.nav-item{role: "presentation", class: ('active' if current_page?(admin_user_edit_history_path()))}
+  -#   = link_to 'Edit History', admin_user_edit_history_path(@user), class: 'nav-link'
+  -# %li.nav-item{role: "presentation", class: ('active' if current_page?(admin_user_locations_path()))}
+  -#   = link_to 'Login Locations', admin_user_locations_path(@user), class: 'nav-link'
```
eanders: Let's go ahead and bring this code over too so we can see who edited the user and where the user logged in from.

tquang1708:
On HMIS the audit button still shows regardless of permission. Should I update that on HMIS too?

eanders:
Sure, just open a new pull request.  Thanks!

eanders (changes requested):
It looks like geocoding isn't working quite correctly.  The warehouse has an additional initializer that may be all it needs. 
`config/initializers/geocode_lite.rb`

If you login and run a delayed job worker, you should see city, region, country data populate in the `login_activities` table.

Also, we should expose the login information as we do in the warehouse under the Edit User pages:
https://qa-warehouse.openpath.host/account/locations

Otherwise, I think this looks great!

tquang1708:
I was able to see the delayed worker do something in regards to the `geocode_lite` initializer, though I'm not sure what that changed (the ip address recorded on my machine is a private IP address). In any case, that seem to be working. I managed to expose the login history information too, so if the geocoding is working as intended then we should be in a good position.

PR #401, opened 2021-01-15
Role editor port

tquang1708:
https://www.pivotaltracker.com/story/show/176488585


tquang1708:
Turns out all we needed was the datatables.min.js file.

In regards to the mixins, in hmis, the mixins are divided up into multiple folders while here it's all stored in the mixins.scss file on cas.

In regards to the checkbox.scss file, what I inserted was basically the diff between the checkbox file on cas and on hmis. Seems like the role table on hmis uses a different checkbox, which explains the extension on the mixins.

In regards to the display.scss file, it helps hide away results for the role table (or basically setting display:none for elements with class 'hide'.) This was necessary for the role table to function, so I don't know whether this is the correct place to put it but the content of the file was not present originally on cas.

PR #404, opened 2021-01-22
Read-only view for COVID assessments within Non-HMIS clients

eanders:


app/models/non_hmis_client.rb
```diff
@@ -36,13 +36,15 @@ def current_covid_assessment
     where(available: true)
   end
 
-  scope :visible_to, -> (user) do
+  scope :visible_to, ->(user) do
     if user.can_edit_all_clients?
       all
+    elsif user.can_view_all_covid_pathways?
+      where(id: NonHmisAssessment.select(:non_hmis_client_id))
```
awisema: Should this have a`type` guard, or is it really can view all assessments?

eanders: So, maybe it should.  The end result is that it is guarded, but in the controller and view.

app/views/non_hmis_clients/assessments/_index.haml
```diff
@@ -16,15 +17,30 @@
...
+                  - if can_manage_identified_clients? || Config.get(:identified_client_assessment) != 'IdentifiedCovidPathwaysAssessment'
+                    %td
+                      - if assessment&.assessment_score&.zero?
+                        0 – Ineligible
+                      - else
+                        = assessment.assessment_score
+                - # If we can't, just show COVID Pathways assessments because those are
+                - # the only assessments that provide sanitized read-only views
+              - else
+                - next unless assessment.type.in?(['DeidentifiedCovidPathwaysAssessment', 'IdentifiedCovidPathwaysAssessment'])
```
awisema: Push this up into the assessment?

awisema (changes requested):

awisema (approved):

PR #405, opened 2021-01-26
Rails 6.1 update

tquang1708:
Attempting to push cas to rails 6.1


tquang1708:
I got all tests to pass on cas so this should be good for deeper review now.

spec/models/message_spec.rb
```diff
@@ -7,6 +7,7 @@
   describe 'When no user is connected to a contact' do
     describe 'Sending a message with deliver_now' do
       it 'creates a message' do 
+        debugger
```
eanders: This should be removed.

spec/models/message_spec.rb
```diff
@@ -23,7 +24,7 @@
       it 'enqueues a job' do 
         expect {
           TestDatabaseMailer.with(email: contact.email).ping.deliver_later
-        }.to have_enqueued_job.on_queue('mailers')
+        }.to have_enqueued_job.on_queue('default')
```
eanders, line 26: I think these should still be in the `mailers` queue.  Maybe there was a change to the way delayed messages get queued?

tquang1708: I think so. It says here that the queue name has been changed.

https://github.com/rails/rails/blob/6-1-stable/actionmailer/CHANGELOG.md

tquang1708:
I attempted to run the broken slack notifier comment in the development environment and it was alright there, so I really don't know why it isn't working in production. For now, I added a safe accessor as a placeholder, but it seems to have run into a different issue.

PR #408, opened 2021-02-03
Show a confirmation dialog on voucher submit if the string has been t…

eanders:
…ranslated.


app/models/voucher.rb
```diff
@@ -51,6 +51,16 @@ def units_including_unavailable
     [unit] + unit.building.units_for_vouchers.to_a
   end
 
+  def self.confirmation_text
+    # NOTE: this must match the string below, but because of the translation
+    # engine's fickleness seems to need to be kept separate.
+    'Please confirm you intend to make vouchers active'
```
awisema: Can we It clearer that a translation makes the text appear? Something like 'Translate to confirm...'?

app/views/vouchers/_form.haml
```diff
@@ -6,7 +6,11 @@
     = render 'voucher', f: f, v: v
   .d-flex
     - if vouchers.count > 0
-      = f.button :submit, value: 'Update Vouchers', class: 'btn btn-primary ml-auto', disabled: @subprogram.closed?
+      - show_confirm = Voucher.confirmation_text != Voucher.translated_confirmation_text
+      - submit_data = {}
+      - if show_confirm
+        - submit_data.merge!({ confirm: Voucher.translated_confirmation_text })
+      = f.button :submit, value: 'Update Vouchers', class: 'btn btn-primary ml-auto', disabled: @subprogram.closed?, data: submit_data
```
awisema, line 13: I thought you were only going to display the confirmation if the user was making vouchers active?

awisema (changes requested):

PR #410, opened 2021-02-05
Generate CE Events from CAS

awisema:


app/models/cas_seeds/match_decision_reasons.rb
```diff
@@ -7,63 +7,66 @@
...
+      ['Client has disappeared', nil],
+      ['SSP CORI', nil],
+      ['HSP CORI', nil],
+      ['Incarcerated', nil],
+      ['Vacancy should not have been entered', nil],
+      ['Client received another housing opportunity', nil],
+      ['Client no longer eligible for match', nil],
+      ['Client deceased', nil],
+      ['Vacancy filled by other client', nil],
+    ].freeze
```
eanders: If every one of these is nil, what's the point?  Is this just place-holder until we know what the second values should be?

app/models/cas_seeds/match_decision_reasons.rb
```diff
@@ -77,49 +80,56 @@ def run!
     end
 
     private def create_other_reason!
-      ::MatchDecisionReasons::Other.all.first_or_create! name: "Other"
+      reason = ::MatchDecisionReasons::Other.all.first_or_create! name: 'Other'
+      reason.update(referral_result: nil)
```
eanders, line 84: With all of these, if these are always create (which I think they are), it will save a query to use the block notation to add the referral_results.

app/models/warehouse/project.rb
```diff
@@ -10,5 +10,25 @@ class Project < Base
 
     belongs_to :organization, class_name: Warehouse::Organization.name, primary_key: [:OrganizationID, :data_source_id], foreign_key: [:OrganizationID, :data_source_id], inverse_of: :projects
 
+    def name
+      self.ProjectName
```
eanders, line 16: I wonder if this should enforce returning "confidential project" if the project is confidential.  I can see a scenario where someone makes the attachment to a program and then the project is later flagged as confidential.

app/models/warehouse/referral_event.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+module Warehouse
+  class ReferralEvent < Base
+    self.table_name = :cas_referral_events
+
+    belongs_to :client, foreign_key: :cas_client_id, inverse_of: :referral_events
+    belongs_to :client_opportunity_match, inverse_of: :referral_event
```
eanders: I'm not sure exactly how this will be handled cross-db, but I think these should get `optional: true`

eanders (changes requested):
This looks generally good, see below for questions/changes.

app/models/cas_seeds/match_decision_reasons.rb
awisema: Yes, I figured we would need to have a discussion on what the disposition of each status would be...

eanders (approved):

PR #419, opened 2021-03-04
Expose client location in an editable way for non-HMIS clients

eanders:


app/controllers/deidentified_clients_controller.rb
```diff
@@ -20,6 +20,7 @@ def create
   end
 
   def update
+    add_shelter_history(clean_params(deidentified_client_params))
```
awisema, line 23: It looks to me like with pushing the client params and clients_path down we could pull update up to `NonHmisClientsController`?

app/views/non_hmis_clients/_shelter_location_form.haml
```diff
@@ -0,0 +1,29 @@
+- if @non_hmis_client.persisted?
+  - url = send("shelter_location_#{client_type}_client_path", @non_hmis_client.id)
```
awisema: I'd rather the url come from a helper method than dynamic programming, but not enough to argue about it.

awisema (approved):

app/controllers/deidentified_clients_controller.rb
eanders: I'm going to leave this now, but I agree, we should eventually clean it up.

PR #423, opened 2021-03-17
Remove expires? from step

awisema:
I think we will also need to run
`ClientOpportunityMatch.open.select { |m| m.current_decision.type == 'MatchDecisions::Four::MatchRecommendationHsa' }.update(shelter_expiration: nil)`
To remove the expiration dates from any current step 3s.


eanders:
Nice! That was way less painful than I expected.

PR #428, opened 2021-04-06
NewLease Match Route

awisema:


app/models/match_decisions/five/five_application_submission.rb
```diff
@@ -0,0 +1,115 @@
...
+    include MatchDecisions::AcceptsDeclineReason
+
+    validate :application_date_present_if_status_complete
+
+    def step_name
+      _('Submit Client Application')
+    end
+
+    def actor_type
+      _('Route Five Shelter Agency')
```
eanders: Good call

app/models/notifications/five/application_submission.rb
```diff
@@ -0,0 +1,23 @@
...
+        create! match: match, recipient: contact
+      end
+    end
+
+    def decision
+      match.five_application_submission_decision
+    end
+
+    def event_label
+      "#{_('Route Five Shelter Agency')} notified of new match"
```
eanders: This looks like a mis-match

app/models/notifications/five/client_agrees.rb
```diff
@@ -0,0 +1,23 @@
...
+        create! match: match, recipient: contact
+      end
+    end
+
+    def decision
+      match.five_client_agrees_decision
+    end
+
+    def event_label
+      "#{_('Route Five Shelter Agency')} notified of new match"
```
eanders: This looks like a mis-match

app/models/rules/holds_voucher.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+    source_match.housing_subsidy_admin_contacts.each do |contact|
+      match.assign_match_role_to_contact(:housing_subsidy_admin, contact)
+    end
+    source_match.shelter_agency_contacts.each do |contact|
+      match.assign_match_role_to_contact(:shelter_agency, contact)
+    end
+    source_match.client_contacts.each do |contact|
+      match.assign_match_role_to_contact(:client, contact)
+    end
```
eanders, line 43: I _think_ this should also copy over:
```
ssp_contacts
hsp_contacts
do_contacts
```

app/views/match_decisions/_five_lease_up.haml
```diff
@@ -0,0 +1,13 @@
...
+      .o-choose.o-choose--flush.o-choose--flush-top
+        .o-choose__choice{ class: ('o-choose__choice--100' unless can_reject_matches?)}
+          %header
+            .o-choose__title
+              .c-choice-icon.c-choice-icon--confirm.mr-4
+              %h3 Client Housed
+          .o-choose__content
+            = render 'match_decisions/client_move_in_details', form: form, submit_label: 'Client Housed'
+
+        = render 'match_decisions/decline_and_cancel_backup_actions', form: form, action_message: 'If the match is declined, the DND will be informed of the decision.'
```
eanders: I think DND should be translated as `_('Route Five DND Contact')` or similar.

app/views/match_decisions/_five_mitigation.haml
```diff
@@ -0,0 +1,26 @@
...
+              .c-choice-icon.c-choice-icon--confirm.mr-4
+              %h3 Mitigation
+          .o-choose__content
+            .form-group.jSelections
+              =form.input :mitigations, as: :check_boxes, collection: @match.match_mitigation_reasons, checked: @match.match_mitigation_reasons.addressed.pluck(:id), label: 'Successfully mitigated', input_html: {disabled: (!@decision.editable?)}
+            .form-actions
+              %button{ class: 'btn btn-success jAccept', data: {submit_param_name: 'decision[status]', submit_param_value: 'accepted'}, disabled: (!@decision.editable?) }
+                %i.icon-checkmark
+                Continue Match
+        = render 'match_decisions/decline_and_cancel_backup_actions', form: form, action_message: 'If the match is declined, the DND will be informed of the decision.'
```
eanders: Same here for DND translation

eanders (commented):
This looks great.  One bit I'd love to have you explore to see how hard it would be, is to translate or pass down contact type names from the route, so in each route we could have a different name for Shelter Agency Contacts or HSA Contacts, hopefully all the way through the app, but generally on the labels on the match page.

app/models/match_routes/five.rb
```diff
@@ -0,0 +1,72 @@
...
+      case contact_type
+      when :dnd_staff_contacts
+        _('Route Five DND Contact')
+      when :housing_subsidy_admin_contacts
+        _('Route Five HSA')
+      when :shelter_agency_contacts
+        _('Route Five Shelter Agency')
+      else
+        super(contact_type)
+      end
```
eanders: This should contain all contact types, missing ssp, hsp, and do.

app/views/contact_manager/_contact_field.haml
```diff
@@ -1,4 +1,7 @@
-- label = object.label_for(contact_type)
+- match_label ||= nil
```
eanders: That's an odd syntax since it will only call the = side of the equation if match_label is nil.  Maybe you meant `defined?`?  Something like `show_link = false unless defined?(show_link)`

app/views/match_contacts/_form.haml
```diff
@@ -1,17 +1,17 @@
 = simple_form_for @match_contacts, url: access_context.match_contacts_path(@match), html: { class: 'current-step-form jContactForm' }, remote: ajax_modal_request? || request.xhr? do |form|
-  = render 'contact_manager/contact_field', contact_type: :shelter_agency_contacts, input_name: 'match_contacts[shelter_agency_contact_ids][]', object: @match_contacts, form: form, current_contact: @current_contact
+  = render 'contact_manager/contact_field', contact_type: :shelter_agency_contacts, input_name: 'match_contacts[shelter_agency_contact_ids][]', object: @match_contacts, match: @match, form: form, current_contact: @current_contact
```
eanders: Maybe for consistency, we should pass in MatchRoute instead of match?  That would mean it could be the same on the program default contacts side, and the contact_field wouldn't need to see if there was a match, just require a route passed in.

app/views/matches/_match_contacts_form.haml
```diff
@@ -9,21 +9,21 @@
         = @match_contacts.errors.join ". "
   .row
     .col-sm-6
-      = render 'contact_manager/contact_field', contact_type: :shelter_agency_contacts, input_name: 'match_contacts[shelter_agency_contact_ids][]', object: @match_contacts, form: form, required: true
+      = render 'contact_manager/contact_field', contact_type: :shelter_agency_contacts, input_name: 'match_contacts[shelter_agency_contact_ids][]', object: @match_contacts, match: @match, form: form, required: true
```
eanders: Same here, route over match if possible.

eanders:
We need a migration for clients at the step that grants vouchers to pre-set their voucher condition.

eanders:
@awisema , can you expose if a client is holding a voucher on the client screen?

PR #430, opened 2021-04-08
Stories/importer adjustments

eanders:


app/models/imported_clients_csv.rb
```diff
@@ -1,40 +1,38 @@
 class ImportedClientsCsv < ApplicationRecord
-
   mount_uploader :file, ImportedClientsCsvFileUploader
 
-
   FORM_TIMESTAMP = 0 # A
   USER_EMAIL = 1 # B
   # HOUSEHOLD_ACKNOWLEDGEMENT = 2 # C - not in use
```
awisema: If these columns are removed from the import, I think we should remove the comments as they refer to incorrect cell names / numbers?

awisema (changes requested):

PR #432, opened 2021-04-09
Syncs ECS code from the warehouse

yakloinsteak:


yakloinsteak:
@eanders When this all passes, I'd like to deploy to qa and then to all of staging. That okay?

eanders:
> @eanders When this all passes, I'd like to deploy to qa and then to all of staging. That okay?

Yes!

yakloinsteak:
@eanders Okay to merge this? I'd like this to be mainstreamed soon.

config/deploy/docker/assets/deploy_tasks.open-path-warehouse.sh
```diff
@@ -2,6 +2,9 @@
 
 set -eo pipefail
 
+echo Storing Themed Maintenance Page
+bundle exec rake maintenance:create
+
```
eanders, line 7: I don't think we're setup to do this in CAS

PR #434, opened 2021-04-14
Adjust pathways assessment permissions

awisema:


app/models/deidentified_client.rb
```diff
@@ -21,6 +21,10 @@ def pathways_enabled?
     Config.get('deidentified_client_assessment').include?('Pathways')
   end
 
+  def assessment_type
```
eanders: Let's call this `assessment_class_name`

app/models/deidentified_covid_pathways_assessment.rb
```diff
@@ -14,4 +14,51 @@ def title
...
+        :shelter_name,
+        :case_manager_contact_info,
+        :mailing_address,
+        :day_locations,
+        :night_locations,
+        :other_contact,
+        :household_size,
+        :pending_subsidized_housing_placement,
+        :pending_housing_placement_type_other,
+        :income_total_annual,
```
eanders: Let's pull out any of these that are shared into a method on non_hmis_assessment so we don't forget to add them in multiple places in the future.

app/models/non_hmis_assessment.rb
```diff
@@ -23,6 +23,18 @@ class NonHmisAssessment < ActiveRecord::Base
     where(type: covid_assessment_types)
   end
 
+  scope :visible_by, ->(user) do
+    where(non_hmis_client_id: NonHmisClient.visible_to(user).select(:id))
+  end
```
eanders, line 29: This is correct.

app/models/non_hmis_assessment.rb
```diff
@@ -23,6 +23,18 @@ class NonHmisAssessment < ActiveRecord::Base
...
+
+  scope :editable_by, ->(user) do
+    where(non_hmis_client: NonHmisClient.editable_by(user).select(:id))
+  end
```
eanders: This needs to inspect the assessment to determine if it was created by someone at your agency.  If so, then you can edit.  (Store agency_id on the assessment)

```
where(
        arel_table[:agency_id].eq(nil).
        or(arel_table[:agency_id].eq(user.agency.id)),
      )
```

app/views/non_hmis_assessments/_common_questions.haml
```diff
@@ -0,0 +1,122 @@
...
+      .c-question__input
+        = form.input :entry_date, label: label, as: :date_picker
+.c-question
+  = form.input :domestic_violence, as: :pretty_boolean, wrapper: :custom_boolean, pre_label: '2B.', label: 'Are you currently fleeing or attempting to flee a domestic violence situation?', hint: 'Examples: sexual assault, stalking, domestic violence'
+.c-question
+  .c-question__wrapper
+    .c-question__number 4D.
+    - label = _('3D. Total # of Boston Homeless Nights')
+    - if disabled
+      = render 'non_hmis_assessments/question_display', value: @assessment.days_homeless_in_the_last_three_years, label: label
```
eanders: Do we actually use this file?  It feels like this should be covered by `@assessment.class.form_fields`

db/migrate/20210416133245_copy_agency_id_into_assessments.rb
```diff
@@ -0,0 +1,7 @@
+class CopyAgencyIdIntoAssessments < ActiveRecord::Migration[6.0]
+  def change
```
eanders: This should be `up`

app/views/non_hmis_assessments/_index.haml
```diff
@@ -0,0 +1,60 @@
...
+              - new_assessment_url = new_polymorphic_path([@non_hmis_client, :non_hmis_assessment])
+            = link_to new_assessment_url, class: 'btn btn-sm btn-secondary' do
+              %span.icon-plus
+              Add New Assessment
+      - if @assessments.any?
+        %table.table
+          %thead
+            %tr
+              %td Date Taken
+              - if can_manage_identified_clients? || Config.get(:identified_client_assessment) != 'IdentifiedCovidPathwaysAssessment'
```
eanders: I don't think this is correct, shouldn't it infer from the assessment based on type? (or maybe from the controller)

app/views/non_hmis_assessments/_index.haml
```diff
@@ -0,0 +1,60 @@
...
+                - if assessment.default?
+                  - url = polymorphic_path([@non_hmis_client], assessment_id: assessment )
+                - else
+                  - if assessment.editable_by?(current_user)
+                    - url = edit_polymorphic_path([@non_hmis_client, :non_hmis_assessment], id: assessment)
+                  - else
+                    - url = polymorphic_path([@non_hmis_client, :non_hmis_assessment], id: assessment)
+                %tr{data: {href: url}}
+                  %td= link_to_if assessment.viewable_by?(current_user), assessment.updated_at, url
+                  - if can_manage_identified_clients? || Config.get(:identified_client_assessment) != 'IdentifiedCovidPathwaysAssessment'
```
eanders: I don't think this is correct, shouldn't it infer from the assessment based on type? (or maybe from the controller)

PR #435, opened 2021-04-16
Fix for names that are too long and the same

eanders:


app/views/reports/match_progress/index.xlsx.axlsx
```diff
@@ -1,7 +1,7 @@
 wb = xlsx_package.workbook
 
-@included_sub_programs.each do |sub_program_id, sub_program_name|
-  wb.add_worksheet(name: sub_program_name[0..30]) do |sheet|
+@included_sub_programs.each.with_index do |(sub_program_id, sub_program_name), i|
+  wb.add_worksheet(name: "#{sub_program_name[0..29]}#{i}") do |sheet|
```
awisema, line 4: I think this assumes that there will never be more than 9 subprograms with long names... Probably OK, but maybe track the names in a hash instead and only add a counter to dups?

eanders: I believe we get 31 characters (maybe even 32) so I think this is good up to 99.

PR #440, opened 2021-04-28
Sync capacity provider logic

yakloinsteak:


config/deploy/docker/vault_poll_state
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env ruby
```
eanders: Why are we deleting this?

yakloinsteak: I was syncing from the warehouse. Maybe it's not over there? I'll bring it back in.

PR #441, opened 2021-04-29
Optionally disable opportunity on cancel

awisema:


db/migrate/20210428205931_add_disable_opportunity_to_match_decisions.rb
```diff
@@ -0,0 +1,5 @@
+class AddDisableOpportunityToMatchDecisions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :match_decisions, :disable_opportunity, :boolean, default: false
+  end
+end
```
eanders, line 5: Why not just put attr_accessor on MatchDecisions::Base?

awisema: I thought it might be useful to be able to preserve this history.

PR #450, opened 2021-05-12
Stories/importer updates

eanders:


app/models/imported_clients_csv.rb
```diff
@@ -64,12 +64,12 @@ def import(agency)
       if assessment.imported_timestamp.nil? || timestamp > assessment.imported_timestamp
         @clients << client
         @touched += 1 if assessment.imported_timestamp.present?
+        binding.pry
```
awisema: Overcommit missed this one!

app/models/imported_clients_csv.rb
```diff
@@ -92,6 +92,7 @@ def import(agency)
           sixty_two_plus: yes_no_to_bool(row[SIXTY_TWO]),
```
awisema, line 91: If we have age, we could compute these directly instead of them sending them to us, yes?

app/models/imported_clients_csv.rb
```diff
@@ -203,6 +206,9 @@ def calculate_dob(row)
     return Date.new(Date.current.year - 62) if yes_no_to_bool(row[SIXTY_TWO])
     return Date.new(Date.current.year - 55) if yes_no_to_bool(row[FIFTY_FIVE])
```
awisema: I think we should we prefer the age computed DOB if we have it...

awisema (changes requested):

app/models/imported_clients_csv.rb
eanders: I'm not sure, the age is technically for HoH, which _should_ be the client in question.  I think I'll leave these as is since they are explicit questions, but use age over them when calculating a fake DOB.  That will capture the responses as uploaded for these age related questions, but will use the provided age for the matching.

awisema (approved):

PR #465, opened 2021-06-21
Maintain client holds voucher from non-hmis assessments

eanders:


app/controllers/non_hmis_assessments_controller.rb
```diff
@@ -73,6 +73,9 @@ def destroy
       assessment_params[:neighborhood_interests] = assessment_params[:neighborhood_interests]&.reject(&:blank?)&.map(&:to_i)
     end
 
+    # Cleanup Vouchers; for now, just check for the word voucher in the response to 3E
+    assessment_params[:have_tenant_voucher] = assessment_params[:pending_housing_placement_type]&.include?('voucher')
```
awisema: `assessment_params[:pending_housing_placement_type]&.downcase` to make the test case in-sensitive?

app/models/match_decisions/record_client_housed_date_housing_subsidy_administrator.rb
```diff
@@ -106,20 +106,20 @@ def holds_voucher?
...
+        match.client.update(holds_voucher_on: Date.current, holds_internal_cas_voucher: true) unless match.has_buildings?
       end
 
       def other_clients_canceled
         match.cancel_opportunity_related_matches!
       end
 
       def completed
-        match.client.update(holds_voucher_on: nil)
+        match.client.update(holds_voucher_on: nil, holds_internal_cas_voucher: nil)
```
awisema, line 117: I think there is a possible problem with clearing `holds_voucher_on` in the step... I don't think the sync from project_client will happen to set it back?

awisema (changes requested):

app/models/match_decisions/record_client_housed_date_housing_subsidy_administrator.rb
eanders: Can you explain a bit more?  This should remove the "holds voucher" when proceeding to the next step, if the non-HMIS client has a voucher, I believe it will be re-added on next sync.

app/controllers/non_hmis_assessments_controller.rb
eanders: 👍

PR #469, opened 2021-06-24
Enable multi-select on cohort rule

eanders:
This is the Cohort equivalent of this: https://github.com/greenriver/boston-cas/pull/464

### DIFF
--- a/app/models/rules/active_in_cohort.rb
+++ b/app/models/rules/active_in_cohort.rb
@@ -11,6 +11,7 @@ class Rules::ActiveInCohort < Rule
 
   def available_cohorts
     return [] unless Warehouse::Base.enabled?
+
     Warehouse::Cohort.visible_in_cas.active.pluck(:id, :name, :short_name).
       map do |id, name, short_name|
         label = short_name.presence || name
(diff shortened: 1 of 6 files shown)


spec/models/rules/active_in_cohort_spec.rb
```diff
@@ -52,7 +58,36 @@
...
+      end
+      it 'does not contain Sue' do
+        expect(clients_that_fit_multi.ids).to_not include sue.id
+      end
+      it 'does not contain Zelda' do
+        expect(clients_that_fit_multi.ids).to_not include zelda.id
+      end
+      it 'does not contain Zelda' do
+        expect(clients_that_fit_multi.ids).to_not include zelda.id
+      end
```
kwiliarty: I think these lines are identical to the previous three.

app/models/rules/active_in_cohort.rb
```diff
@@ -19,19 +20,24 @@ def available_cohorts
...
+    # Note, we need to jump through a few hoops since variable wasn't designed as a json field
+    available_cohorts.select { |id, _| id.to_s.in?(value_as_array(value)) }.map(&:last).join(' or ') || value
   end
 
-  def clients_that_fit(scope, requirement, opportunity)
+  def clients_that_fit(scope, requirement, _opportunity)
     if Client.column_names.include?(:active_cohort_ids.to_s)
       if requirement.positive
-        where = 'active_cohort_ids @> ?'
+        where = 'active_cohort_ids @> ANY(ARRAY [?]::jsonb[])'
```
kwiliarty, line 30: I'm not familiar with this syntax. Can you tell me what it's doing?

eanders: The question mark will be replaced with the value when dropped into the where statement down below.  They @> is postgres json query syntax for "in" (https://www.postgresql.org/docs/9.5/functions-json.html.)  So, force the string provided into an postgres array, and see if any of the provided values are in the active_cohort_ids column.

kwiliarty: It was the `@>` that was new to me. Thanks!

kwiliarty (approved):
This looks good to me.

PR #471, opened 2021-06-28
Creates polymorphic housing media links

kwiliarty:


kwiliarty:
Reviewing this myself I think I can and should just take out the heading and subheading pieces. We needed that for the building attributes because they were more complicated.

kwiliarty:
Need to add the links on the building show page

kwiliarty:
Open in new tab

kwiliarty:
Add building and unit links on the match page, fix building edits

kwiliarty:
Hard code heading, but make it translatable.

kwiliarty:
Show url in edit table

app/views/housing_media_links/new.haml
```diff
@@ -0,0 +1,13 @@
+-# Partial: used polymorphically for building and unit attributes
+-#   Arguments:
+-#     housingable: currently a building or unit object
+-#     link: a HousingMediaLink object
+
+= content_for :modal_title do
+  Add Photo or Video Link
```
eanders: Let's translate this `= _('Add Photo or Video Link')`

kwiliarty:
> Need to add the links on the building show page

And actually, it is helpful here to have a parameterized heading after all, because we need to suppress the heading in some contexts, and this is one of them.

kwiliarty:
Adding the links to the match pages was slightly tricky, so that was the only part that wasn't fast.

PR #472, opened 2021-06-28
Uses a more accurate test of whether an email body is in HTML format

kwiliarty:


eanders (approved):
@kwiliarty , very nice.  Thanks for finding this. Will merge after tests pass and a quick run-through.

PR #473, opened 2021-06-28
Removes obsolete unit_attributes references

kwiliarty:


kwiliarty:
I'm pretty sure that these references to `unit_attributes` are obsolete. I missed them the first time around, but we may as well clean them up now, I think.

PR #474, opened 2021-07-05
Stories/building pictures 178525861

kwiliarty:
Adds validation for URL field, but breaks out of modal. Do you know a way to stay within the model through the validation process?


eanders:
@kwiliarty, you'd probably have to switch to using `remote: true` on the form for them to show up within the modal.  In this case, I don't think it's worth the effort right now as it would also require setting up the JS to maintain the rest of the page after the save.  Something like https://github.com/greenriver/hmis-warehouse/blob/production/app/views/health/goals/create.js.coffee

In the mean-time, can you fix the merge conflict with pre-release? 
Thanks!

kwiliarty:
Ah, I missed that you added the `respond_with` change. Sorry for the confusion. I'm not sure what the `respond_with` change is doing, so I'm not sure how/whether to incorporate it into the latest create method with validation:

```
    1    def create
>>  2 <<<<<<< HEAD
    3     @media_link = @housingable.housing_media_links.where(media_link_params).first_or_create
    4     respond_with @media_link, location: edit_polymorphic_path(@housingable)
>>  5 =======
    6     @media_link = @housingable.housing_media_links.where(media_link_params).first_or_initialize
    7     if @media_link.valid?
    8       @media_link.save
    9       redirect_to edit_polymorphic_path(@housingable)
   10     else
   11       render :new
   12     end
>> 13 >>>>>>> 6dc7542d (Validates URL (but breaks out of modal))
   14   end
```

PR #480, opened 2021-07-22
Stories/match contact type 178711392

kwiliarty:
This is still a draft because I have a question about line 108 in `app/models/match_contacts.rb`. Is it okay to make that modification? I couldn't tell whether client contacts had been omitted intentionally or not. Including it certainly makes things easier to include them, if it's legit.


eanders:
> This is still a draft because I have a question about line 108 in `app/models/match_contacts.rb`. Is it okay to make that modification? I couldn't tell whether client contacts had been omitted intentionally or not. Including it certainly makes things easier to include them, if it's legit.

I don't see any issue with having that there.  The only use of it is in listing the contacts on the match in `ClientOpportunityMatch#current_contact_titles`, which doesn't appear to be an issue.

eanders (approved):

kwiliarty:
> > This is still a draft because I have a question about line 108 in `app/models/match_contacts.rb`. Is it okay to make that modification? I couldn't tell whether client contacts had been omitted intentionally or not. Including it certainly makes things easier to include them, if it's legit.
> 
> I don't see any issue with having that there. The only use of it is in listing the contacts on the match in `ClientOpportunityMatch#current_contact_titles`, which doesn't appear to be an issue.

Excellent. :-)

PR #485, opened 2021-08-17
Stories/simplify non hmis assessments

eanders:


app/controllers/non_hmis_clients_controller.rb
```diff
@@ -159,20 +144,12 @@ def sorter
...
-  def build_assessment
-    assessment_type.constantize.new
-  end
-
   def load_client
     @non_hmis_client = client_source.find(params[:id].to_i)
     @assessment = load_assessment || @non_hmis_client.current_assessment
   rescue Exception
     client = NonHmisClient.find params[:id].to_i
     case client.type
```
awisema: Only one `when` left here? replace with
```redirect_to polymorphic_path([action_name.to_sym, :imported_client], id: params[:id]) if client.type == 'ImportedClient'```?

File: app/models/deidentified_client_assessment.rb [hunk 2]
```diff
@@ -27,4 +32,61 @@ def viewable_by?(user)
...
+        view_helper.concat 'View'
+        view_helper.concat view_helper.content_tag(:span, nil, class: 'icon-arrow-right2 ml-2')
+      end
+    end
+    delete_link = view_helper.link_to(url_helpers.deidentified_client_path(client), method: :delete, data: { confirm: 'Would you really like to delete this Non-HMIS client?' }, class: ['btn', 'btn-sm', 'btn-danger']) do
+      view_helper.concat(view_helper.content_tag(:span, nil, class: 'icon-cross'))
+      view_helper.concat(view_helper.content_tag(:span, 'Delete'))
+    end
+    row << client_link if user.can_view_some_clients? && client.client
+    row << delete_link if user.can_manage_deidentified_clients? && ! client.involved_in_match?
```
awisema: I think the client guard should just be for the def of `client_link` `and `delete_link`? Otherwise the columns won't line up if the permission is true, but the client guard is false...
```
row << client_link if user.can_view_some_clients?
row << delete_link if user.can_manage_deidentified_clients?
```

app/models/deidentified_client_assessment.rb
```diff
@@ -27,4 +32,61 @@ def viewable_by?(user)
...
+      'Client Identifier',
+      'Agency',
+      'Assessment Score',
+      'Assessment Date',
+      'Cohorts',
+      'Days Homeless in the Last 3 Years',
+      'Available?',
+    ]
+    columns << '' if user.can_manage_deidentified_clients?
+    columns << 'CAS Client' if user.can_view_some_clients?
```
awisema: These are in the opposite order in the client table row below...

app/models/deidentified_client_assessment.rb
```diff
@@ -27,4 +32,61 @@ def viewable_by?(user)
...
+      checkmark(client.available?),
+    ]
+
+    client_link = if client.client
+      view_helper.link_to(url_helpers.client_path(client.client), class: 'btn btn-secondary btn-sm d-inline-flex align-items-center') do
+        view_helper.concat 'View'
+        view_helper.concat view_helper.content_tag(:span, nil, class: 'icon-arrow-right2 ml-2')
+      end
+    end
+    delete_link = view_helper.link_to(url_helpers.deidentified_client_path(client), method: :delete, data: { confirm: 'Would you really like to delete this Non-HMIS client?' }, class: ['btn', 'btn-sm', 'btn-danger']) do
```
awisema: I think this needs to be `delete_link = unless client_involved_in_match? ...`

app/models/deidentified_client_assessment.rb
awisema: This and above applies to the other assessments as well.

app/models/identified_client_assessment.rb
```diff
@@ -27,4 +27,63 @@ def viewable_by?(user)
...
+      'First Name',
+      'Agency',
+      'Assessment Score',
+      'Assessment Date',
+      'Cohorts',
+      'Days Homeless in the Last 3 Years',
+      'Available?',
+    ]
+    columns << '' if user.can_manage_identified_clients?
+    columns << 'CAS Client' if user.can_view_some_clients?
```
awisema: The order of these two columns is inconsistent among the assessment types... Dunno if we care...

app/views/non_hmis_assessments/edit.haml
```diff
@@ -1,3 +1,3 @@
-= render 'non_hmis_assessments/form',
-  disabled: false,
-  action_text: 'Update'
+= render 'non_hmis_assessments/form', disabled: false, action_text: 'Update'
+
+-# = render "#{@non_hmis_client.model_name.element}s/#{@assessment.model_name.element}", disabled: false
```
awisema: Clean up required?

awisema (changes requested):

awisema (approved):

PR #487, opened 2021-08-20
New on-demand capacity provider, etc.

yakloinsteak:
New capacity provider for on-demand and capacity provider for scheduled tasks


yakloinsteak:
Going to wait and see output of `rake cas:add_missing_tie_breakers`, which runs on the hour. If that's good, then I think this can be merged.

PR #489, opened 2021-08-25
Notifies slack on unexpected nil match (179058521)

kwiliarty:


app/controllers/match_list_base_controller.rb
```diff
@@ -207,6 +213,13 @@ def available_programs
   # via targeted search
   private def visible_match_ids
     contact = current_user.contact
+    # We are trying to figure out why the match side of a client/match pair is sometimes nil.
+    # We haven't been able to reproduce the problem so we are at least going to try to notify it.
+    contact.client_opportunity_match_contacts.each do |contact_match_pair|
+      if contact_match_pair.match.nil?
+        @notifier.ping "The match for the contact/match pair <id: #{contact_match_pair.id}> with user <id: #{current_user.id}> is nil" if @send_notifications
```
eanders: Let's make this
```
@notifier.ping "The match #{contact_match_pair.match_id} for the contact/match pair <id: #{contact_match_pair.id}> with user <id: #{current_user.id}> is nil" if contact_match_pair.match.nil? && @send_notifications
```

kwiliarty: ✅

PR #502, opened 2021-09-14
Prevent changes to Pathways V3 assessements for a while

eanders:


app/models/concerns/pathways_version_three_calculations.rb
```diff
@@ -43,6 +43,48 @@ def assessment_type_options
...
+      lock_days = Config.get(:lock_days)
+      return if lock_days.zero?
+      return if in_lock_grace_period?
+
+      self.locked_until = Date.current + lock_days.days
+    end
+
+    def in_lock_grace_period?
+      return true if lockable_fields.empty?
+      return false if locked_until.blank?
```
awisema, line 77: I don't know if these should be true or false, but it seems like they should be the same? Maybe false? You can't be in the grace period if there isn't one, or there are no fields?

app/models/non_hmis_assessment.rb
```diff
@@ -255,6 +255,22 @@ def genders
...
+  def locked?
+    false
+  end
+
+  # Do nothing, override as necessary
+  def lock
+  end
+
+  def in_lock_grace_period?
+    true
```
awisema, line 271: Should match whatever is above?

app/models/concerns/pathways_version_three_calculations.rb
eanders: If `locked_until` is blank, we can't be in the grace period or it prevents setting the lock date in `lock`

app/models/non_hmis_assessment.rb
eanders: This helps prevent setting `locked_until` in the situation where it's not relevant because we never set `locked_until` if we're in the grace period.

awisema (approved):

PR #505, opened 2021-09-15
Rule to match specific assessment type

eanders:


app/models/non_hmis_assessment.rb
```diff
@@ -58,6 +58,20 @@ def self.limited_assessment_types
...
+      merge(IdentifiedPathwaysAssessment.new.for_matching).
+      merge(DeidentifiedPathwaysAssessment.new.for_matching).
+      merge(IdentifiedCovidPathwaysAssessment.new.for_matching).
+      merge(DeidentifiedCovidPathwaysAssessment.new.for_matching).
+      merge(IdentifiedPathwaysVersionThree.new(assessment_type: :pathways_2021).for_matching).
+      merge(IdentifiedPathwaysVersionThree.new(assessment_type: :transfer_assessment).for_matching).
+      merge(DeidentifiedPathwaysVersionThree.new(assessment_type: :pathways_2021).for_matching).
+      merge(DeidentifiedPathwaysVersionThree.new(assessment_type: :transfer_assessment).for_matching).
+      freeze
+  end
```
awisema, line 73: I'd like to turn this into something like
```
{}.tap do |hash|
  known_assessment_types.each do |clazz|
    hash.merge(clazz.new.for_matching)
  end
end
```
But, I'm not sure what to do with this `assessment_type` parameter on PathwaysV3...

app/models/rules/unsheltered.rb
```diff
@@ -0,0 +1,19 @@
...
+      scope.where(unsheltered: requirement.positive)
+    else
+      raise RuleDatabaseStructureMissing.new("clients.unsheltered missing. Cannot check clients against #{self.class}.")
+    end
+  end
+
+  def always_apply?
+    true
+  end
+end
```
awisema: Didn't you decide to replace this with `EnrolledInSo`?

awisema (changes requested):

app/models/rules/unsheltered.rb
eanders: Yep, guess I accidentally added the file.  Removed.

app/models/non_hmis_assessment.rb
eanders: Yeah, that's exactly why I ended up where it is.
It is useful for the class to know which slugs/titles to show for the rule, but because the latest iteration requires an "assessment_type" it all broke down.  Happy to entertain other suggestions.

awisema: Maybe add a `for_matching` param that default to false, and then `clazz.new(for_matching: true).for_matching`?

PR #506, opened 2021-09-15
Displays links to active matches on separate lines

kwiliarty:


kwiliarty:
This turned out to be very minor, indeed, but it's still an improvement, I believe.

app/views/qualified_opportunities/index.haml
```diff
@@ -7,7 +7,7 @@
     %p
       #{@client.full_name} is already active in at least one match:
       - @client.active_matches.each do |match|
-        %br
+        %p
```
eanders, line 10: This is fine, but I actually think the issue is just that the `=` on the line below the `%br` is indented.  Can you test and see if just outdenting that to the same depth as the %br fixes it? (Adding p tags seems extreme and may bring unwanted padding along for the ride.)

kwiliarty: Outdenting doesn't help. If I remove the p-tag in line 7 then it does. Everything then gets centered, though.

eanders: Thanks for testing.  p tag it is!

PR #507, opened 2021-09-17
Stories/site config for name visibility 178865169

awisema:


app/models/client_opportunity_match.rb
```diff
@@ -271,18 +270,19 @@ def show_client_info_to? contact
...
     end
   end
 
   def client_info_approved_for_release?
-    if match_route.class.name.in?([ 'MatchRoutes::Default' ])
-      return shelter_agency_approval_or_dnd_override? && client&.has_full_housing_release?
+    if match_route.class.name.in?(['MatchRoutes::Default'])
+      shelter_agency_approval_or_dnd_override? && client&.has_full_housing_release?
     else
-      client&.has_full_housing_release? || false
```
eanders, line 285: I love that we had the false in here. I guess that catches a nil, but....

PR #510, opened 2021-10-13
Client address report

awisema:


app/views/reports/housed_addresses/index.haml
```diff
@@ -0,0 +1,31 @@
...
+        %thead
+          %tr
+            - headers.each do |header|
+              %th= header
+        %tbody
+          - @addresses.pluck(columns).each do |row|
+            %tr
+              - row.each do |value|
+                - if !!value == value # Check for boolean
+                  %td= value ? 'Y' : 'N'
```
eanders: Let's use one of the various available methods in `application_helper`  Maybe, `yes_no`.

PR #512, opened 2021-10-20
(179839355) Implement 'latest-*' to support ECR lifecycle policy

andyzito:
Same as https://github.com/greenriver/hmis-warehouse/pull/1881


.github/workflows/ecs.yml
```diff
@@ -109,12 +109,33 @@ jobs:
...
+          HAS_DJ=true
           echo pushing dj
           docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG--dj
         else
           echo not pushing dj
         fi
 
+        echo Pushing 'latest' tags
+        BRANCH=${GITHUB_REF#refs/heads/}
+        if [[ "$BRANCH" =~ ^(production|staging|sandbox|demo|main|release-*|deploy-.*)$ ]]; then
```
yakloinsteak: I think this approach made sense for HNMI, but over here, I think modifying the config/deploy/docker/lib scripts (maybe just deployer.rb) would be better. After that code waits for the build, and before the RollOut class is called, we know the environment and client name, so that should be enough to tag. Let me know, @andyzito if you want to discuss. This will apply to the warehouse as well.

yakloinsteak (changes requested):
Probably need something different for cas/warehouse.

.github/workflows/ecs.yml
andyzito: @yakloinsteak So to confirm, Open Path (both CAS and Warehouse) don't use the GHA workflow to push images?

yakloinsteak:
They do, but that's not the place where we know that a particular image is
deployed to a particular environment. Let's talk tomorrow.

andyzito:
@eanders This is now identical to the one from warehouse (https://github.com/greenriver/hmis-warehouse/pull/1881). Assuming @yakloinsteak has no objections, you can merge at your leisure.

eanders:
Do we not need the gems here too?

yakloinsteak:
Add the gems, yes.

andyzito:
Sure, we can include those here, although I will note that the gem issues were to do with my environment more than the functionality in this PR. Still, it's good to get them into the bundle file. @eanders I'm going to recommend that you run the command to update `Gemfile.lock` on your end, because I don't think my ruby+gem environment is in a very good place right now.

yakloinsteak (approved):

PR #516, opened 2021-10-27
Stories/funding prioritization scheme

eanders:


app/models/match_prioritization/assessment_score_funding_tie_breaker.rb
```diff
@@ -0,0 +1,23 @@
...
+module MatchPrioritization
+  class AssessmentScoreFundingTieBreaker < Base
+    def self.title
+      'Assessment Score with Funding Date Tie Breaker'
+    end
+
+    def self.prioritization_for_clients(scope, match_route:) # rubocop:disable Lint/UnusedMethodArgument
+      scope.where.not(c_t[:assessment_score].eq(nil)).
+        order(c_t[:assessment_score].desc).
+        order(c_t[:financial_assistance_end_date].desc)
```
awisema: So, I _think_ this should be `.asc`?

app/models/match_prioritization/days_homeless_last_three_years_assessment_date.rb
```diff
@@ -0,0 +1,23 @@
...
+module MatchPrioritization
+  class DaysHomelessLastThreeYearsAssessmentDate < Base
+    def self.title
+      'Days homeless in the last three years with earliest assessment date tie breaker'
+    end
+
+    def self.prioritization_for_clients(scope, match_route:) # rubocop:disable Lint/UnusedMethodArgument
+      scope.order(Arel.sql(c_t[:days_homeless_in_last_three_years].desc.to_sql + ' NULLS LAST')).
+        order(Arel.sql(c_t[:entry_date].asc.to_sql + ' NULLS LAST')).
+        order(c_t[:tie_breaker].asc)
```
awisema, line 16: Does the funding tiebreaker need this coin-flip as well?

app/models/non_hmis_client.rb
```diff
@@ -236,11 +241,7 @@ def set_project_client_fields(project_client)
     project_client.sro_ok = current_assessment&.sro_ok
     project_client.evicted = current_assessment&.evicted
     project_client.ssvf_eligible = current_assessment&.ssvf_eligible || false
-    project_client.holds_voucher_on = if current_assessment&.have_tenant_voucher
-      current_assessment&.entry_date
-    else
-      nil
-    end
+    project_client.holds_voucher_on = (current_assessment&.entry_date if current_assessment&.have_tenant_voucher)
```
awisema: I'm not sure this is equivalent -- the old code would set `holds_voucher_on` to `nil` if it was set and `have_tenant_voucher` was false... I think this will leave it set.

spec/models/client_spec.rb
```diff
@@ -57,8 +59,41 @@
         expect( Client.prioritized(route, Client.all).pluck(:id, :assessment_score, :days_homeless) ).to eq Client.order(assessment_score: :desc, rrh_assessment_collected_at: :desc).pluck(:id, :assessment_score, :days_homeless)
       end
     end
+
+    context 'when prioritized by AssessmentScore with funding tie breaker' do
+      let(:priority) { create :priority_assessment_score_funding_tie_breaker }
+      let(:route) { create :default_route, match_prioritization: priority }
+      before :each do
+        # set the two top scores to the same thing, make sure we get the client with the latest financial_assistance_end_date
+        # as the highest priority
```
awisema: I _thought_ this was supposed to go the other way, but I'm unclear on the spec.

awisema (changes requested):

app/models/match_prioritization/days_homeless_last_three_years_assessment_date.rb
eanders: no harm

app/models/non_hmis_client.rb
eanders: Oh, this was rubocop.  I didn't remember this change.

eanders: I believe wrapping it in () would return nil if false, since the `if` occurs within the `()`.  But I think this is easier to read:
```
project_client.holds_voucher_on = nil # reset to nil
project_client.holds_voucher_on = current_assessment&.entry_date if current_assessment&.have_tenant_voucher
```

awisema: I definitely prefer your suggestion, as that seems a dark corner case of the language.

awisema (approved):

PR #532, opened 2021-12-07
More Pathways Rules

awisema:


app/models/rules/majority_unsheltered.rb
```diff
@@ -0,0 +1,15 @@
+###
+# Copyright 2016 - 2021 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+class Rules::MajorityUnsheltered < Rule
+  def clients_that_fit(scope, requirement, opportunity)
+    if Client.column_names.include?(:majority_sheltered.to_s)
+      scope.where(majority_sheltered: ! requirement.positive)
```
eanders, line 10: Does this not need to ignore nil?

eanders (approved):

app/models/rules/enrolled_in_hmis_project_type.rb
```diff
@@ -0,0 +1,74 @@
...
+    type_clauses = if requirement.positive
+      clauses.join(' or ')
+    else
+      clauses.join(' and ')
+    end
+
+    scope.where(type_clauses)
+  end
+
+  private def clause(column_name, positive)
```
eanders: Are you actually using this?

app/models/rules/enrolled_in_hmis_project_type.rb
```diff
@@ -0,0 +1,74 @@
...
+      a_t[type_columns[project_type.to_sym]].eq(requirement.positive).to_sql
+    end.compact
+
+    type_clauses = if requirement.positive
+      clauses.join(' or ')
+    else
+      clauses.join(' and ')
+    end
+
+    scope.where(type_clauses)
```
eanders: Does this need an `Arel.sql(type_clauses)` Looks like it's just passing plain sql to where.

app/models/rules/majority_unsheltered.rb
awisema: I don't think so -- SQL NULL is neither true nor false.

app/models/rules/enrolled_in_hmis_project_type.rb
awisema: Nope! Removed.

app/models/rules/enrolled_in_hmis_project_type.rb
awisema: Fixed.

eanders (approved):

PR #534, opened 2021-12-07
Generic tie_breaker_date

awisema:


app/models/non_hmis_client.rb
```diff
@@ -191,6 +191,7 @@ def set_project_client_fields(project_client) # rubocop:disable Naming/AccessorM
     project_client.entry_date = current_assessment&.entry_date
     project_client.rrh_assessment_collected_at = current_assessment&.entry_date
     project_client.financial_assistance_end_date = current_assessment&.financial_assistance_end_date
+    project_client.tie_breaker_date = current_assessment&.financial_assistance_end_date
```
eanders: This will only be `financial_assistance_end_date` if it's a transfer assessment, otherwise, `entry_date`

eanders: Also, `assessment_score` needs to be set to the days homeless in the past 3 years, if it's a pathways v3 (and not transfer) assessment.

eanders (changes requested):
Two tweaks to the non-hmis -> project client sync.

eanders (approved):
Just waiting on tests

PR #535, opened 2021-12-08
Stories/new nights fields

eanders:


app/controllers/concerns/program_permissions.rb
```diff
@@ -31,7 +31,7 @@ module ProgramPermissions
     end
 
     private def check_edit_permission!
-      not_authorized! unless can_edit_programs? || (can_edit_assigned_programs? && @subprogram.editable_by?(current_user))
+      not_authorized! unless can_edit_programs? || (can_edit_assigned_programs? && @subprogram&.editable_by?(current_user))
```
awisema, line 34: Seems off that `@subprogram` would ever be nil, but it fails safe...

app/models/non_hmis_client.rb
```diff
@@ -247,6 +247,13 @@ def set_project_client_fields(project_client) # rubocop:disable Naming/AccessorM
     project_client.holds_voucher_on = current_assessment&.entry_date if current_assessment&.have_tenant_voucher
     project_client.enrolled_in_es = current_assessment&.enrolled_in_es || false
     project_client.enrolled_in_so = current_assessment&.enrolled_in_so || false
+    # if we have breakdowns of sheltered vs unsheltered we can set majority_sheltered
+    # otherwise, just leave it as nil
+    if current_assessment&.homeless_nights_sheltered.present? && current_assessment&.homeless_nights_unsheltered.present? && current_assessment&.additional_homeless_nights_sheltered.present? && current_assessment&.additional_homeless_nights_unsheltered.present?
+      sheltered = current_assessment.homeless_nights_sheltered + current_assessment.additional_homeless_nights_sheltered
+      unsheltered = current_assessment.homeless_nights_unsheltered + current_assessment.additional_homeless_nights_unsheltered
+      project_client.majority_sheltered = sheltered > unsheltered
```
awisema: Didn't you set it to `nil` when they were equal when coming from Clarity?

awisema (changes requested):

app/controllers/concerns/program_permissions.rb
eanders: Yeah, we've been seeing odd can't do fu on nil, so I just spammed them around a bit.

awisema (approved):

PR #543, opened 2021-12-22
Stories/tc hat

eanders:


app/models/rules/challenge.rb
```diff
@@ -0,0 +1,44 @@
...
+    [
+      'No Income/Set Income',
+      'Domestic Violence',
+      'No rental history',
+      'Bad Credit',
+      'Eviction History',
+      'Misdemeanor',
+      'Felony Conviction /Criminal History',
+      'Sex Offender',
+      'Meth production conviction',
```
awisema, line 22: The inconsistent capitalization worries me that they may change it in the assessment. Maybe title case them and down case the json values? (There is a note for that on the Warehouse side as well).

app/models/rules/strength.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def available_strengths
+    [
+      'Reliable Vehicle',
+      'Regular Income',
+      'Employable Skills',
+      'No Disabling Conditions',
+      'Good Credit',
+      'Recent Positive Rental History',
```
awisema, line 19: Downcase here too?

awisema (changes requested):
Should their be a `app/views/admin/rules/challenges/_challenge.haml`?

awisema (approved):

PR #545, opened 2022-01-07
Limit shelter agency contacts to active

awisema:


eanders (approved):
Looks great!  I'll deploy to QA so we can test

PR #552, opened 2022-01-24
I is variable, so trim the name appropriately

eanders:


app/views/reports/match_progress/index.xlsx.axlsx
```diff
@@ -1,7 +1,7 @@
 wb = xlsx_package.workbook
 
 @included_sub_programs.each.with_index do |(sub_program_id, sub_program_name), i|
-  wb.add_worksheet(name: "#{sub_program_name[0..29].gsub(/['\*\/\\\?\[\]\:]/, '-')}#{i}") do |sheet|
+  wb.add_worksheet(name: "#{sub_program_name[0..(30 - i)].gsub(/['\*\/\\\?\[\]\:]/, '-')}#{i}") do |sheet|
```
awisema: As `i` is a counter, and we are creating a string from it, I think, it needs to be:
```suggestion
  wb.add_worksheet(name: "#{sub_program_name[0..(30 - i.to_s.length)].gsub(/['\*\/\\\?\[\]\:]/, '-')}#{i}") do |sheet|
```
?

awisema (changes requested):

app/views/reports/match_progress/index.xlsx.axlsx
eanders: Yep, that makes more sense.

PR #553, opened 2022-01-25
Move contact editing to users

eanders:


app/controllers/contacts_controller.rb
```diff
@@ -82,47 +48,45 @@ def update_matches
...
+    params.require(:contact).permit(
+      :first_name,
+      :middle_name,
+      :last_name,
+      :email,
+      :phone,
+      :cell_phone,
+      :role,
+    )
+  end
```
awisema: I think this is dead code.

app/views/contacts/index.html.haml
```diff
@@ -1,53 +1,47 @@
...
-      %span.icon-plus
-      Add a Contact
-
+%p Contacts are now managed through their associated user accounts.
 - @prompt = t '.search_prompt'
 .row
   .col-sm-4
     = render 'search_form'
 :ruby
-  columns = ['last_name', 'first_name', 'middle_name', 'email', 'phone', 'role']
```
awisema, line 13: I wonder if we should keep role around for the legacy contacts?

config/routes.rb
```diff
@@ -106,7 +106,7 @@ def manage_matches
 
   end
 
-  resources :contacts, except: :show do
+  resources :contacts, except: [:show, :edit, :update, :create, :destroy] do
```
awisema:
```suggestion
  resources :contacts, only: [:index] do
```

awisema (changes requested):

app/views/contacts/index.html.haml
eanders: I'm removing it from the view since you can't set it anymore.  I'm guessing they should all be blank anyway and the role should come from the contact's participation on the match.  If someone asks to have it back, we can just re-enable it and add it to the user form.

app/controllers/contacts_controller.rb
eanders: yep

PR #555, opened 2022-01-27
Expose cancel/park/backup on acknowledge screen; fix tab defaults

eanders:


app/views/match_decisions/_backup_tab.haml
```diff
@@ -1,5 +1,7 @@
+- active_tab = '' unless defined?(active_tab)
```
awisema, line 1: Should this ever be set? I see `render ..., active_tab: 'active'` for the other ones, but not for `backup_tab`. If this is for consistency, that is fine.

awisema (approved):

app/views/match_decisions/_backup_tab.haml
eanders: yep, just for consistency so we when we later add a new tab we're not confused.

PR #563, opened 2022-02-22
Include match note in email

awisema:


app/controllers/admin/configs_controller.rb
```diff
@@ -33,10 +33,20 @@ def update
...
     end
 
+    def include_note_in_email_options
+      {
+        'Never include match notes in email notifications' => nil,
+        'Allow match notes in email notifications (default to no)' => false,
+        'Allow match notes in email notifications (default to yes)' => true,
+      }
+    end
+    helper_method :include_note_in_email_options
```
eanders: This should be a method on config.rb

app/views/match_decisions/_match_recommendation_shelter_agency_fields.haml
```diff
@@ -16,6 +16,8 @@
         Client provided a
         = _(@match.client.housing_release_status)
 
+- unless Config.get(:include_note_in_email_default).nil?
```
eanders, line 19: I don't think, you want a required boolean.

awisema: Maybe I was being lazy, but this is 3-valued logic... if it is set to `true` it defaults to including the note in the email, if it is `false` it defaults to not being included, but shows the checkbox to allow it to be included. And if it is `nil`, then it it is not included, and the checkbox is hidden.

eanders: My mistake, I tagged the wrong line.  I meant to tag the input on line 20

eanders (approved):

PR #564, opened 2022-02-24
Shelter agency user can decline on steps 3, 4, and 5 of the default r…

eanders:
…oute, admins can approve/reject declines


app/controllers/match_decisions_controller.rb
```diff
@@ -137,86 +133,83 @@ def update
...
     end
+    redirect_to access_context.match_decision_path(@match, @decision, redirect: 'true')
   end
 
   private
 
-    def find_match!
-      @match = match_scope.find params[:match_id]
-    end
+  def find_match!
```
awisema: In addition to being un-indented, these should get their own `private` declaration/

app/models/match_decisions/approve_match_housing_subsidy_admin.rb
```diff
@@ -121,15 +139,19 @@ def canceled
...
+        end
+      end
     end
     private_constant :StatusCallbacks
 
     private def note_present_if_status_declined
-      if note.blank? && status == 'declined'
-        errors.add :note, 'Please note why the match is declined.'
-      end
+      errors.add :note, 'Please note why the match is declined.' if note.blank? && status == 'declined'
```
awisema: Should this also be true for `shelter_declined`?

app/models/match_decisions/five/five_screening.rb
```diff
@@ -36,8 +36,8 @@ def statuses
       {
         pending: 'Pending',
         expiration_update: 'Pending',
-        mitigation_required: "#{_('Mitigation required')}",
-        mitigation_not_required: "#{_('No mitigation required')}",
+        mitigation_required: _('Mitigation required').to_s,
+        mitigation_not_required: _('No mitigation required').to_s,
```
awisema: I think the `to_s` is redundant?

docker/docker-compose.yml
```diff
@@ -21,6 +21,7 @@ x-app: &app
     PSQL_HISTFILE: /app/log/.psql_history
     EDITOR: vi
     CHROMEDRIVER_VERSION: 79.0.3945.36
+    BUNDLE_FORCE_RUBY_PLATFORM: true
```
awisema, line 25: This line blows up for me -- something about how true isn't a valid type for a docker environment variable. Does it not do it for you?

awisema (approved):

docker/docker-compose.yml
eanders: It works for me, but I did have to `docker-compose build`.  Let me know if that doesn't fix it for you, I have a less good solution we can try.

app/models/match_decisions/approve_match_housing_subsidy_admin.rb
eanders: This doesn't appear to be connected as one would expect.  Declines don't actually have a place to input a note.  But it works as expected.

PR #568, opened 2022-03-03
Non-HMIS client import changes for HAT

awisema:


eanders:
@awisema, it looks like this includes some of some other branch?  Does the importer really need app/controllers/client_contact_controller.rb and app/mailers/client_mailer.rb?  I can't seem to get the PR changes to a reasonable state.  Is b4cd53d the only commit that matters?  Maybe we can rebuild the branch off of `release-15`?

awisema:
Abandoning this PR in favor of a clean one in https://github.com/greenriver/boston-cas/pull/569

PR #574, opened 2022-03-21
New rules for TC HAT

eanders:


app/models/non_hmis_client.rb
```diff
@@ -279,15 +279,32 @@ def populate_project_client(project_client)
     project_client.strengths = current_assessment&.strengths&.reject(&:blank?)
     project_client.challenges = current_assessment&.challenges&.reject(&:blank?)
     project_client.open_case = current_assessment&.tc_hat_client_history&.include?('open_case')
-    project_client.foster_care = current_assessment&.foster_care
     project_client.housing_for_formerly_homeless = current_assessment&.housing_preferences&.include?('with_formerly_homeless')
     project_client.drug_test = current_assessment&.drug_test
```
awisema: You moved this down, so it can be removed, I think?

app/models/non_hmis_client.rb
```diff
@@ -279,15 +279,32 @@ def populate_project_client(project_client)
...
+      :living_wage,
+      :need_daily_assistance,
+      :full_time_employed,
+      :can_work_full_time,
+      :willing_to_work_full_time,
+      :rrh_successful_exit,
+      :lifetime_sex_offender,
+      :th_desired,
+      :drug_test,
+      :employed_three_months,
```
awisema, line 300: Duplicated from above

app/models/rules/currently_fleeing.rb
```diff
@@ -0,0 +1,25 @@
...
+    raise RuleDatabaseStructureMissing.new("clients.#{column} or clients.#{column_2} missing. Cannot check clients against #{self.class}.") unless Client.column_names.include?(column.to_s) && Client.column_names.include?(column_2.to_s)
+
+    c_t = Client.arel_table
+    if requirement.positive
+      scope.where(
+        c_t[column].eq(requirement.positive).
+        and(c_t[column_2].gteq(3.months.ago.to_date)),
+      )
+    else
+      scope.where(column => requirement.positive).
```
awisema: I'd find
```suggestion
      scope.where(column => false).
```
more readable.

awisema (changes requested):

PR #577, opened 2022-03-25
Stories/shelter agency contact emails

eanders:


app/views/notifications_mailer/_shelter_agency_contacts.text.haml
```diff
@@ -1,2 +1,2 @@
 #{_('Case Workers')} for this match:
-#{@match.shelter_agency_contacts.map{|c| "#{c.name.presence} #{c.phone.presence if c.phone != 'x'}"}.join("\n")}
\ No newline at end of file
+#{@match.shelter_agency_contacts.map { |c| [c.name.presence, c.email, (c.phone.presence if c.phone != 'x')].map(&:presence).compact.join(' ') }.join("\n") }
```
awisema: I there any reason to apply `presence` twice?
```suggestion
#{@match.shelter_agency_contacts.map { |c| [c.name, c.email, (c.phone if c.phone != 'x')].map(&:presence).compact.join(' ') }.join("\n") }
```

awisema (changes requested):

app/views/notifications_mailer/_shelter_agency_contacts.text.haml
eanders: Makes sense.  I was converting from a string to an array with a join, missed that we didn't need that in both places.  Thanks!

awisema (approved):

PR #581, opened 2022-03-30
Prevent default contacts from being added for Deactivated Accounts

gigxz:
**Problem**: Deactivated users continue to be added as default contacts for new matches. The deactivated users weren't being notified, but it was causing confusion to have them on new matches.

**Solution:** Don't add inactive users as default contacts to new matches. For matches where the deactivated users are already listed as contacts, they'll remain there (with the existing caution tooltip indicating that they are inactive).


eanders (approved):
Looks good to me!

app/models/client_opportunity_match.rb
```diff
@@ -739,25 +741,25 @@ def assign_match_role_to_contact role, contact
   end
 
   private def add_default_dnd_staff_contacts!
-    Contact.where(user_id: User.dnd_initial_contact.select(:id)).each do |contact|
+    Contact.where(user_id: User.dnd_initial_contact.select(:id)).preload(:user).each do |contact|
```
awisema, line 744: Maybe use `Contact#active_contacts` to avoid needing to retrieve the user entirely?

awisema (changes requested):

app/models/client_opportunity_match.rb
gigxz: For this one specifically, we actually already know that all the contacts are active because we modified the `User#dnd_initial_contact` scope to limit to active users. (Unlike other scopes `sub_program.dnd_staff_contacts`, `opportunity.housing_subsidy_admin_contacts` and so on, those might contain inactive users.)

So, even without the `Contact#active_contacts` scope, you're right that we really don't need to do the user retrieval here.. I think the purpose of keeping it was just to use the same helper method `assign_match_role_to_contact` to keep things consistent. Was that your thought when we were pairing on this @eanders?  If we really want to get rid of this query, we could. One other approach we discussed was doing a join and merge on `User.active` for each contact scope, which would allow us to remove the user retrieval from `assign_match_role_to_contact`.

eanders: @sandy, the issue is that in `assign_match_role_to_contact` we ask the user if they are active (so we can prevent adding them if they aren't active.)  If we don't preload, we end up getting an N+1 in `assign_match_role_to_contact`.  In practice, we're never adding more than 2 or 3 contacts per type, so the difference doesn't really matter.  I say leave it.

PR #588, opened 2022-04-11
Fix log stream search to be much, much better

andyzito:
See https://github.com/greenriver/hmis-warehouse/pull/2133

@eanders This includes inserting the log stream URL into error notifications. Please trigger an error from CAS QA when you get a chance to confirm that is working.


eanders:
@andyzito , this looks good, except it's not attaching the log URL to errors thrown when browsing.  Do you remember where that code gets injected?

andyzito:
@eanders Looks like release-19 got redeployed -- sent this code back up, looks like the URL is coming through now.

PR #592, opened 2022-04-27
Add match group prioritization scheme

gigxz:


app/models/match_prioritization/match_group.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module MatchPrioritization
+  class MatchGroup < Base
+    def self.title
+      'Match Group'
+    end
+
+    def self.prioritization_for_clients(scope, match_route:) # rubocop:disable Lint/UnusedMethodArgument
+      scope.order(Arel.sql(c_t[:match_group].asc.to_sql + ' NULLS LAST')).
+        order(Arel.sql(c_t[:entry_date].asc.to_sql + ' NULLS LAST')).
```
eanders, line 15: We should double check entry_date is the date you are expecting. I don't believe that is currently flowing from the warehouse.  Pretty sure this should be `rrh_assessment_collected_at`

eanders (approved):

app/models/match_prioritization/match_group.rb
gigxz: I updated the warehouse exporter to set `entry_date` equal to the same value as `rrh_assessment_collected_at`, as you suggested on slack
https://github.com/greenriver/hmis-warehouse/pull/2173/files

Although one difference is that `rrh_assessment_collected_at` column is a datetime and `entry_date` column is a date. That may be an unintended consequence. All assessments from one day will be considered equally prioritized, and vispdat may end up being used as a tie breaker more often.

eanders: I believe that should be fine, good catch though!

PR #596, opened 2022-05-11
Ea/ruby upgrade

eanders:


andyzito (approved):

Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'composite_primary_keys', '~> 12.0'
 gem 'csv', '>= 1.0.2' # support for bom|utf-8 in ruby 2.5
 gem 'order_as_specified'
 gem 'with_advisory_lock'
-gem 'nokogiri', '>= 1.13.2' # CVE-2021-30560
+gem 'nokogiri', '>= 1.13.4' # CVE-2021-30560
```
awisema: I think this comment should be removed 1.13.4 doesn't go with that CVE.
Also, Should we set this to 1.13.6, which is for [CVE-2022-29181](https://nvd.nist.gov/vuln/detail/CVE-2022-29181) and is what is in the `Gemfile.lock`?

awisema (changes requested):

awisema (approved):

PR #600, opened 2022-05-20
Add admin functions to stalled matches

eanders:


awisema (approved):
I ran through some matches, and this is working as I would expect.

PR #602, opened 2022-05-25
[PVT-181020089] Port deployment improvements from warehouse

andyzito:
[PVT#181020089](https://www.pivotaltracker.com/story/show/181020089)

See also https://github.com/greenriver/hmis-warehouse/pull/2234, which is a few followup changes to Warehouse that are also included in this PR.

This PR includes all of the deployment improvements we made as part of the referenced Pivotal story, as well as the follow up changes referenced above. 

I was able to successfully deploy to QA. We should test thoroughly in staging to ensure that all clients receive the correct assets.


.github/workflows/asset_compilation.yml
```diff
@@ -0,0 +1,135 @@
...
+        WAREHOUSE_DATABASE_HOST: postgres
+        WAREHOUSE_DATABASE_PASS: postgres
+        WAREHOUSE_DATABASE_USER: postgres
+        WAREHOUSE_DATABASE_DB: warehouse_db
+        HEALTH_DATABASE_DB: health_db
+        REPORTING_DATABASE_DB: reporting_db
+        DATABASE_DB_TEST: boston_cas_test
+        DATABASE_WAREHOUSE_DB_TEST: test_hmis_warehouse
+        HOSTNAME: openpath.host
+        PORT: 80
```
andyzito, line 116: @eanders Was hoping it would be easy for you to eyeball which of these we actually need and delete the rest?

eanders: I don't think any of those are in use for asset recompilation but I'm unsure if you could run `bundle exec bin/compile_assets.rb` without them.  Maybe try getting rid of everything but `HOSTNAME`?

eanders: Oh, sorry, just which ones are relevant to CAS.

.github/workflows/asset_compilation.yml
```diff
@@ -0,0 +1,135 @@
...
+        REPORTING_DATABASE_HOST: postgres
+        REPORTING_DATABASE_PASS: postgres
+        REPORTING_DATABASE_USER: postgres
+        WAREHOUSE_DATABASE_ADAPTER: postgis
+        WAREHOUSE_DATABASE_HOST: postgres
+        WAREHOUSE_DATABASE_PASS: postgres
+        WAREHOUSE_DATABASE_USER: postgres
+        WAREHOUSE_DATABASE_DB: warehouse_db
+        HEALTH_DATABASE_DB: health_db
+        REPORTING_DATABASE_DB: reporting_db
```
eanders: None of these should be necessary.
```suggestion
```

.github/workflows/asset_compilation.yml
```diff
@@ -0,0 +1,135 @@
...
+        DATABASE_DB_TEST: boston_cas_test
+        DATABASE_WAREHOUSE_DB_TEST: test_hmis_warehouse
```
eanders: Shouldn't need this either
```suggestion
```

.github/workflows/asset_compilation.yml
andyzito: Yup, thank you 🙏

eanders (approved):
This looks great and works well!  Thanks @andyzito!

PR #607, opened 2022-06-02
Pin alpine to 3.15

gigxz:
Do we want to pin alpine at 3.15 for cas, like we did for the warehouse?
Seems like 3.16 is working fine, but I just noticed the inconsistency.


gigxz:
oh I think we would want to do https://github.com/greenriver/boston-cas/blob/f936b5cf68f40a45c8e4a75b988d127c76e58ce1/config/deploy/docker/assets/Dockerfile.open-path-cas.pre-cache#L4 too

PR #608, opened 2022-06-02
Pin alpine to 3.15 for pre-cache

gigxz:


gigxz:
I don't know if this file or https://github.com/greenriver/hmis-warehouse/blob/production/config/deploy/docker/assets/Dockerfile.open-path-cas.pre-cache is used

eanders:
Definitely used, that's part of the deployment world.

gigxz:
But I mean we have two nearly-identical `Dockerfile.open-path-cas.pre-cache` files in both repos? Should we be keeping them up to date and matching?  Looks like there are several changes that have been made in one file but not the other throughout the past year

eanders:
I believe the idea was to keep the entire `config/deploy/docker` directories in sync, but there has definitely been drift.  I'll add that to @andyzito's queue.

andyzito:
@eanders Can I ask again if it's actually needed to keep these changes synced? It's unclear to me what the benefit is of having the files in the opposing repo, you have to move changes between repos anyway. Seems like it would be easier to just keep changes in sync between the individual warehouse and cas files in their own repos.

eanders:
I'm happy to keep them separate.  @yakloinsteak ?

yakloinsteak:
I think separate is right. The apps don't have the same set of dependencies, though I imagine cas is a subset of the warehouse perhaps? I had a goal of reducing the number of Dockerfiles but that might be off topic. Each type of container can have its own entrypoint and/or command and healthcheck provided at run time.

PR #611, opened 2022-06-07
Add lease end date to housed addresses report

gigxz:
Screen Shot 2022-06-07 at 9 52 42 AM

### DIFF
--- a/app/controllers/reports/housed_addresses_controller.rb
+++ b/app/controllers/reports/housed_addresses_controller.rb
@@ -77,8 +77,8 @@ module Reports
         'Last Name' => { query: c_t[:last_name], display_type: :text },
         'Housing Address' => { query: md_b_t[:address], display_type: :address },
         _('Did you or this client use external software to help with housing?') => { query: md_b_t[:external_software_used], display_type: :boolean },
-        'Lease Start Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt.to_date } },
-        'Lease End Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt.to_date >> 12 } },
+        'Lease Start Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt&.to_date } },
+        'Lease End Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt.to_date.present? && dt.to_date >> 12 } },
         'Match ID' => { query: com_t[:id], display_type: :integer },
       }.freeze
     end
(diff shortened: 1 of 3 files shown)


app/controllers/reports/housed_addresses_controller.rb
```diff
@@ -72,7 +77,8 @@ def column_types
         'Last Name' => { query: c_t[:last_name], display_type: :text },
         'Housing Address' => { query: md_b_t[:address], display_type: :address },
         _('Did you or this client use external software to help with housing?') => { query: md_b_t[:external_software_used], display_type: :boolean },
-        'Date Housed' => { query: md_b_t[:client_move_in_date], display_type: :text },
```
gigxz, line 75: I renamed the "Date Housed" column to "Lease Start Date" for consistency, because that's what we call it everywhere else.

eanders: Thanks!

app/controllers/reports/housed_addresses_controller.rb
```diff
@@ -72,7 +77,8 @@ def column_types
...
+        'Lease Start Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt.to_date } },
```
eanders: does this blow up if `dt` is empty?  maybe?
```suggestion
        'Lease Start Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt&.to_date } },
```

app/controllers/reports/housed_addresses_controller.rb
```diff
@@ -72,7 +77,8 @@ def column_types
...
+        'Lease End Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt.to_date >> 12 } },
```
eanders:
```suggestion
        'Lease End Date' => { query: md_b_t[:client_move_in_date], display_type: :text, display_transform: ->(dt) { dt.to_date.present? && dt.to_date >> 12 } },
```

eanders (approved):
Couple small edge cases I _think_ are worth catching.

app/controllers/reports/housed_addresses_controller.rb
gigxz: it does but `client_move_in_date` is always present https://github.com/greenriver/boston-cas/blob/4fdd885f542778ae22b49b61f6068b1e271eb6c3/app/controllers/reports/housed_addresses_controller.rb#L83

gigxz: but I can update it anyway

eanders: thanks.  I thought that might be the case, but hadn't tracked it that far back.

PR #615, opened 2022-06-10
Add 'Multiple Match Disclaimer' translation

awisema:
Add a translation for the text describing multiple matches.

If the disclaimer is not set. use the existing language. Used 'Multiple Match Disclaimer' as the translation key instead of the existing text for brevity/readability.


app/views/matches/_details.html.haml
```diff
@@ -15,9 +15,15 @@
             .icon-checkmark
           .c-tag__content
             %span= match.overall_status[:name]
-  - if match.opportunity.multiple_active_matches?
+  - if true # match.opportunity.multiple_active_matches?
```
eanders: Was this debugging?

eanders (approved):
Looks good except the `if true`

app/views/matches/_details.html.haml
awisema: Oops! Yes, fixed.

PR #620, opened 2022-06-23
Ea/multi match priority

eanders:


app/models/client_opportunity_match.rb
```diff
@@ -280,14 +280,20 @@ def show_client_info_to? contact
...
-      current_decision != send(match_route.initial_decision)
-    else
-      true
+    match_route.on_or_after_first_client_step?(current_decision)
+  end
+
+  # Preload initialized_decisions
+  def first_client_decision
+    decision = initialized_decisions.detect do |d|
+      d.instance_of?(match_route)
```
awisema: Should this be:
```suggestion
      d.instance_of?(match_route.first_client_step.constantize)
```
?

app/models/match_routes/base.rb
```diff
@@ -69,6 +73,19 @@ def initial_decision
       raise NotImplementedError
     end
 
+    def first_client_step
+      self.class.match_steps.keys.first
```
awisema, line 74: Maybe?
```suggestion
      self.class.first_client_step
```

awisema (changes requested):

awisema (approved):

PR #623, opened 2022-06-27
Update SH XLSX import

awisema:


app/models/deidentified_clients_xlsx.rb
```diff
@@ -65,17 +65,20 @@ def skip?(row, index)
...
+    result[:mental_health_problem] = yes_no_to_bool(client, :mental_health_problem, row[:mental_health_problem])
     result.delete(:occurrences_of_homelessness)
     result[:days_homeless] = convert_to_number(client, :days_homeless, row[:days_homeless])
     result[:family_member] = yes_no_to_bool(client, :family_member, row[:family_member])
-    result[:disabling_condition] = yes_no_to_bool(client, :disabling_condition, row[:disabling_condition])
-    result[:is_currently_youth] = yes_no_to_bool(client, :is_currently_youth, row[:is_currently_youth])
     result[:sixty_plus] = yes_no_to_bool(client, :sixty_plus, row[:sixty_plus])
-    result[:required_number_of_bedrooms] = convert_to_number(client, :required_number_of_bedrooms, row[:required_number_of_bedrooms])
+    result[:is_currently_youth] = yes_no_to_bool(client, :is_currently_youth, row[:is_currently_youth])
+    result[:calculated_chronic_homelessness] = yes_no_to_bool(client, :chronic_homeless, row[:chronic_homeless]) ? 1 : 0
```
eanders, line 79: This looks you're converting a boolean into a number so it can be converted back to a boolean, is that right?

eanders: oh, never mind.  you are converting to a boolean, then picking 1 or 0.

db/schema.rb
```diff
@@ -947,6 +947,7 @@
     t.integer "tc_hat_mobile_home"
     t.integer "tc_hat_total_housing_rank"
     t.integer "days_homeless"
+    t.boolean "pregnancy_status", default: false
```
eanders, line 950: Are you missing the migration?

eanders (approved):

PR #626, opened 2022-06-30
Move all continue match buttons to a sigle partial; expose top/not to…

eanders:
…p priority on matches

### DIFF
--- a/app/models/match_routes/base.rb
+++ b/app/models/match_routes/base.rb
@@ -80,7 +80,15 @@ module MatchRoutes
     end
 
     def on_or_after_first_client_step?(current_decision)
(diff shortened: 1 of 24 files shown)


app/models/match_routes/base.rb
```diff
@@ -80,7 +80,15 @@ def first_client_step
     end
 
     def on_or_after_first_client_step?(current_decision)
-      self.class.match_steps[current_decision.class.name] >= first_client_step_number
+      on_first_client_step?(current_decision) || after_first_client_step?(current_decision)
+    end
```
awisema, line 84: Maybe move this to `ClientOpportunityMatch` to minimize the route API? I don't think it really matters tho...

awisema (approved):

PR #630, opened 2022-07-10
[PVT-182502615] JSON logging

andyzito:
https://github.com/greenriver/hmis-warehouse/pull/2298

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -56,7 +56,6 @@ module BostonCa
         request_time: Time.current,
         application: Rails.application.class,
         server_protocol: event.payload[:server_protocol],
-        forwarded_for: event.payload[:forwarded_for],
         host: event.payload[:host],
         remote_ip: event.payload[:remote_ip],
         ip: event.payload[:ip],
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,4 +1,5 @@
 require_relative 'boot'
+require_relative "../app/logger/log_formatter.rb"
 
 require 'rails/all'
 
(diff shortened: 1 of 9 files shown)


config/application.rb
```diff
@@ -47,18 +48,30 @@ class Application < Rails::Application
...
     config.lograge.enabled = true
+    config.lograge.logger = ActiveSupport::Logger.new(STDOUT)
+    config.lograge.formatter = Lograge::Formatters::Json.new
+    config.lograge.base_controller_class = ['ActionController::Base']
     config.lograge.custom_options = ->(event) do
       {
+        request_time: Time.current,
+        application: Rails.application.class,
         server_protocol: event.payload[:server_protocol],
-        forwarded_for: event.payload[:forwarded_for],
```
eanders, line 53: @andyzito, is it possible this is needed to get the actual IP address?  I have a recollection we needed something special, but maybe it was only necessary before we moved to ECS.

eanders: This seems to be handled with `x_forwarded_for` and appears both are `nil` in our current deployment setup.

andyzito:
@eanders It's unclear to me where we're at here -- is this working? What issue appeared with `forwarded_for` that prompted you to remove it? Do those adjustments need to be backported to warehouse?

eanders:
@andyzito , nothing to see here, I think.  You'd removed the `forwarded_for: event.payload[:forwarded_for],`  I added it back in, deployed and tested to see if it made a difference.  It didn't, so I re-removed it.  It doesn't exist in the warehouse.

andyzito:
@eanders Ah, I think I see -- forwarded_for is added in the application_controller along with the other at-request-time values. Yeah, I think we're good here.

PR #637, opened 2022-07-26
Work around race condition where deleted clients get matched - 182760774

gigxz:
1. Every 5 minutes, delete any `ClientOpportunityMatch` that points to a deleted client
2. If someone pulls up the Opportunity page that has a match afflicted by this race condition (in the 5min before it is removed), hide the match from the page and make sure the app doesn't blow up


gigxz:
Build is failing due to vulnerability, fixed with https://github.com/greenriver/boston-cas/pull/638

PR #647, opened 2022-09-14
Make address required on match route 4

gigxz:


eanders (approved):
I'm not 100% sure about removing `match.opportunity.voucher.unit.blank?` but it seems to meet the request, so let's leave it out unless they say otherwise.

PR #650, opened 2022-09-22
New routes are disabled by default

eanders:


db/schema.rb
```diff
@@ -274,8 +274,8 @@
     t.boolean "enrolled_in_rrh", default: false
     t.boolean "enrolled_in_psh", default: false
     t.boolean "enrolled_in_ph", default: false
-    t.boolean "majority_sheltered"
```
awisema, line 277: Not sure why these move around, but, there don't seem to be any changes here.

awisema (approved):

db/schema.rb
eanders: Yeah, that was just me running migrations.

PR #653, opened 2022-10-05
Support multiple race and gender values

gigxz:
Update to 2020 HUD spec changes.


db/migrate/20221005133804_add_race_gender_columns_to_client.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_column :project_clients, :no_single_gender, :boolean, default: false
+    add_column :project_clients, :transgender, :boolean, default: false
+    add_column :project_clients, :questioning, :boolean, default: false
+
+    add_column :clients, :female, :boolean, default: false
+    add_column :clients, :male, :boolean, default: false
+    add_column :clients, :no_single_gender, :boolean, default: false
+    add_column :clients, :transgender, :boolean, default: false
+    add_column :clients, :questioning, :boolean, default: false
+  end
```
gigxz, line 26: Should I remove the old race and gender columns and drop the Race and Gender tables here as well? Should I add as a second migration? Not sure what we typically do when removing tables/columns. @eanders

app/views/clients/_demographics.haml
```diff
@@ -6,23 +7,24 @@
...
           %span.c-card__content-block-text.col-md-12.col-lg-6.c-card__content-block-text--secondary= "(#{date_format(@client.date_of_birth)})" if @client.date_of_birth
         .col-md-12.col-lg-6.c-card__content-block
           %h5 Gender
-          %span.c-card__content-block-text= @client.gender.try(:text)
+          %span.c-card__content-block-text= find_and_preserve(@client.gender_descriptions.join(',<br>')).presence || not_specified
+
         .col-md-12.col-lg-6.c-card__content-block
           %h5 Race
-          %span.c-card__content-block-text= @client.race.try(:text)
+          %span.c-card__content-block-text= find_and_preserve(@client.race_descriptions.join(',<br>')).presence || not_specified
```
gigxz, line 18: Looks like this. I also added "not specified" to all fields.

Screen Shot 2022-10-05 at 2 35 45 PM

Screen Shot 2022-10-05 at 2 40 08 PM

eanders: This looks good.  Thanks

db/migrate/20221005133804_add_race_gender_columns_to_client.rb
eanders: yes, let's drop those columns and tables.  We'll also need to update the gender rules.

eanders (changes requested):
Can you track down any usages of the gender and race columns?
Here's one:
https://github.com/greenriver/boston-cas/blob/production/app/views/match_client_details/show.html.haml
We'll need to update all of those to the new version.

gigxz:
This has been updated @eanders and should be ready for another review.
- fixed all usages of Race/Gender
- collect multiple races/genders for identified clients, and export them in csv download
- update ProjectClient correctly with race/gender for non hmis clients
- collect multiple genders for household members in Pathways 3
- remove gender and race db tables

app/models/identified_client.rb
```diff
@@ -133,8 +136,8 @@ def download_data
         middle_name,
         last_name,
         date_of_birth,
-        Gender.find_by(numeric: gender)&.text,
-        Race.find_by(numeric: race)&.text,
+        gender_descriptions.join(', '),
+        race_descriptions.join(', '),
```
gigxz, line 140: This is for the identified clients CSV download. @eanders do you think this is okay, or should we have separate columns per value? I don't believe this gets imported anywhere.

db/migrate/20221006175635_drop_race_gender_tables.rb
```diff
@@ -0,0 +1,21 @@
...
+      t.timestamps null: false
+    end
+
+    drop_table :genders do |t|
+      t.integer :numeric
+      t.string :text
+      t.timestamps null: false
+    end
+  end
+end
```
gigxz, line 21: @eanders let me know if you want me to combine these all into one migration

app/models/identified_client.rb
eanders: this looks good.

app/views/clients/_client_list.haml
```diff
@@ -63,7 +63,7 @@
                   unknown
                 = client_data_quality client, :date_of_birth
             %td
-              = client.gender.try(:text)
+              = find_and_preserve(client.gender_descriptions(brief: true).join(',<br>'))
```
eanders, line 66: huh, who knew!

db/migrate/20221005190528_remove_old_race_gender.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveOldRaceGender < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :project_clients, :primary_race, :string
+    remove_column :project_clients, :secondary_race, :string
+    remove_column :project_clients, :gender, :integer
+    remove_column :clients, :race_id, :integer
+    remove_column :clients, :gender_id, :integer
+    remove_column :clients, :gender_other, :string
```
eanders: I think we should leave these temporarily.  If we remove them the two deployments need to go in lock-step or the CAS sync process will start bombing out as soon as we've deployed one or the other.  If we leave them, deploy both, then, remove them, that should maintain the dependency.

db/migrate/20221006175635_drop_race_gender_tables.rb
eanders: nope, multiple is fine.

eanders (changes requested):
Let's remove that one migration, then I think it looks great.  Thanks!

db/migrate/20221005190528_remove_old_race_gender.rb
gigxz: Ah right, updated.

PR #654, opened 2022-10-06
Support deactivating Units

gigxz:
* Expose edit/deactivate/delete/restore on Units page https://boston-cas.dev.test/units
* Expose edit/deactivate/delete/restore on Building page https://boston-cas.dev.test/buildings/1
* Expose deactivate/delete/restore on Edit page https://boston-cas.dev.test/units/10/edit


app/controllers/units_controller.rb
```diff
@@ -67,76 +75,85 @@ def update
...
-        :name,
-        :available,
-        :building_id,
-        :elevator_accessible,
-        requirements_attributes: [:id, :rule_id, :positive, :variable, :_destroy]
-      )
-    end
+  def unit_scope
+    Unit
+  end
```
eanders: usually this would be `unit_source` and `unit_scope` would be something like `unit_source.visible_to(current_user)`.  Given that Unit is only one word, and the scope doesn't include a scope, I'd probably just remove this method and use `Unit` instead as is being done on line 125.

config/routes.rb
```diff
@@ -114,6 +114,7 @@ def manage_matches
     end
   end
   resources :units, except: :show, concerns: [:restorable] do
+    post :deactivate
```
eanders, line 117: do we need a restore route here too?

eanders (approved):
Looks great.  There's a ton going on in there.

config/routes.rb
gigxz: We already had a restore route, that was previously used for Delete, so I reused that one instead

app/controllers/units_controller.rb
gigxz: Yeah agreed. This diff is mostly rubocop fixes. I will update though

PR #657, opened 2022-10-12
Filter matches by contact

gigxz:
Please review:
* As a non-admin user, I can filter by contact type ("Shelter Agency Contact: Me")
* As an admin user, I can...
    * Filter by contact type ("Has Shelter Agency Contact")
    * Filter by user ("Contact: Foo Bar")
    * Filter by both user and contact type ("Shelter Agency Contact: Foo Bar")

### DIFF
--- a/app/controllers/active_matches_controller.rb
+++ b/app/controllers/active_matches_controller.rb
@@ -19,11 +19,10 @@ class ActiveMatchesController < MatchListBaseController
     @current_step = params[:current_step]
     @current_program = params[:current_program]
     @current_contact_type = params[:current_contact_type]&.to_sym
-    if current_user.can_view_all_matches?
-      @current_filter_contact = params[:current_filter_contact]
-      @current_filter_contact = @current_filter_contact.to_i if @current_filter_contact.present?
+    @current_filter_contact = if current_user.can_view_all_matches?
+      params[:current_filter_contact].to_i if params[:current_filter_contact].present?
     elsif @current_contact_type.present?
-      @current_filter_contact = current_user.contact&.id
+      current_user.contact&.id
     end
     @matches = filter_by_step(@current_step, @matches)
     @matches = filter_by_route(@current_route, @matches)
@@ -36,7 +35,7 @@ class ActiveMatchesController < MatchListBaseController
       order(sort_matches)
     @column = sort_column
     @direction = sort_direction
-    @active_filter = @current_step.present? || @current_program.present?
+    @active_filter = [@current_step, @current_program, @current_contact_type, @current_filter_contact].map(&:presence).any?
     @types = MatchRoutes::Base.match_steps
 
     @page_size = 25
--- a/app/controllers/active_matches_controller.rb
+++ b/app/controllers/active_matches_controller.rb
@@ -18,9 +18,17 @@ class ActiveMatchesController < MatchListBaseController
     @matches = match_scope
     @current_step = params[:current_step]
     @current_program = params[:current_program]
+    @current_contact_type = params[:current_contact_type]&.to_sym
+    if current_user.can_view_all_matches?
+      @current_filter_contact = params[:current_filter_contact]
+      @current_filter_contact = @current_filter_contact.to_i if @current_filter_contact.present?
+    elsif @current_contact_type.present?
+      @current_filter_contact = current_user.contact&.id
+    end
     @matches = filter_by_step(@current_step, @matches)
     @matches = filter_by_route(@current_route, @matches)
     @matches = filter_by_program(@current_program, @matches)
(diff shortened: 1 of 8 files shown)


app/controllers/active_matches_controller.rb
```diff
@@ -18,9 +18,17 @@ def index
     @matches = match_scope
     @current_step = params[:current_step]
     @current_program = params[:current_program]
+    @current_contact_type = params[:current_contact_type]&.to_sym
+    if current_user.can_view_all_matches?
+      @current_filter_contact = params[:current_filter_contact]
+      @current_filter_contact = @current_filter_contact.to_i if @current_filter_contact.present?
```
gigxz: I really wanted `params[:current_filter_contact]&.to_i` to return `nil` if the param was missing (or an empty string, I guess) but it was returning `0`. Ugly solution :\

app/controllers/match_list_base_controller.rb
```diff
@@ -168,6 +168,16 @@ def available_programs
   end
   helper_method :available_programs
 
+  def available_contact_types
+    MatchContacts.input_names
```
gigxz: Is this the right set?

app/controllers/match_list_base_controller.rb
```diff
@@ -168,6 +168,16 @@ def available_programs
...
+  end
+  helper_method :available_contact_types
+
+  def available_contacts
+    Contact.active_contacts
```
gigxz: Is this the right set? I thought about unioning the scopes for available contacts of each type, but that seems like potentially unnecessary complexity

app/controllers/active_matches_controller.rb
```diff
@@ -18,9 +18,17 @@ def index
...
+    elsif @current_contact_type.present?
+      @current_filter_contact = current_user.contact&.id
+    end
```
eanders: Maybe:
```suggestion
    @current_filter_contact = if current_user.can_view_all_matches?
      params[:current_filter_contact].to_i if params[:current_filter_contact].present?
@current_filter_contact.present?
    elsif @current_contact_type.present?
      current_user.contact&.id
    end
```
That _should_ set @current_filter_contact = nil if there's no `params[:current_filter_contact]` 
Also, instance variables default to nil instead of throwing an error when accessed if they haven't been set, so if you like the other set better, you could remove line 23.

app/controllers/match_list_base_controller.rb
eanders: I believe so, though since you are only using it once, I'd probably not bother with the controller method and helper method, just shove `MatchContacts.input_names` in the view.

app/controllers/match_list_base_controller.rb
eanders: I believe so, though since you are only using it once, I'd probably not bother with the controller method and helper method, just shove `Contact.active_contacts` in the view.

app/controllers/active_matches_controller.rb
```diff
@@ -18,9 +18,17 @@ def index
     @matches = match_scope
     @current_step = params[:current_step]
     @current_program = params[:current_program]
+    @current_contact_type = params[:current_contact_type]&.to_sym
```
eanders, line 21: This all needs to get ported into `closed_matches_controller.rb` as well.

eanders, line 21: `@active_filter` needs to include the new fields in both `closed_matches_controller.rb` and `active_matches_controller.rb` so the filter reflects that it's active when things are chosen.

app/views/active_matches/index.html.haml
```diff
@@ -35,15 +35,17 @@
...
+      - contact_type = if @current_contact_type.present? then Contact::label_for(@current_contact_type)&.gsub("Contacts", "Contact") else 'Contact' end
+      - contact_type = "Has #{contact_type}" if contact_type.present? && !contact_name.present?
+      - program_filters = [step_name, program_name].compact.join(' at ')
+      - contact_filters = [contact_type, contact_name].compact.join(': ')
       %p.client__current-sort-filter
         Limited to:
-        %strong= limits.join(' at ')
-
-
+        %strong= simple_format([program_filters, contact_filters].filter(&:presence).join(',<br />'), {}, wrapper_tag: 'span')
```
eanders: This also needs to happen in `closed_matches/index.html.haml`

eanders (changes requested):
This looks great!  Just a few more places that it needs to be added to, and some minor cleanup.

gigxz:
Thanks @eanders!! Feedback addressed

eanders (approved):
I need to test locally, but this looks great.  Thanks @gigxz

PR #658, opened 2022-10-13
Columns for open and closed match count

gigxz:
Is this going to be super slow in prod? If so we could pre calculate the counts per contact in the controllers

https://boston-cas.dev.test/contacts
https://boston-cas.dev.test/admin/users

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -100,7 +100,6 @@ Layout/DotPosition:
     - 'app/models/cas_seeds/client_shelter_agency_contact.rb'
     - 'app/models/cas_seeds/opportunity_housing_subsidy_admin_contact.rb'
     - 'app/models/concerns/related_default_contacts.rb'
-    - 'app/models/contact.rb'
     - 'app/models/identified_client.rb'
     - 'app/models/program.rb'
     - 'app/models/sub_program.rb'
@@ -1348,7 +1347,6 @@ Naming/PredicateName:
   Exclude:
     - 'spec/**/*'
     - 'app/controllers/concerns/match_building_and_unit.rb'
-    - 'app/models/contact.rb'
     - 'app/models/match_events/note.rb'
     - 'app/models/match_progress_updates/base.rb'
     - 'app/models/project_client.rb'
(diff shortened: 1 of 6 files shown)


.rubocop_todo.yml
```diff
@@ -100,7 +100,6 @@ Layout/DotPosition:
     - 'app/models/cas_seeds/client_shelter_agency_contact.rb'
     - 'app/models/cas_seeds/opportunity_housing_subsidy_admin_contact.rb'
     - 'app/models/concerns/related_default_contacts.rb'
-    - 'app/models/contact.rb'
```
eanders: Pretty sure this didn't get fixed.  Let's leave the todos here unless we fix them.

app/models/contact.rb
```diff
@@ -98,4 +98,16 @@ def self.contact_type_for(input_name)
   def phone_for_display
     phone.presence || cell_phone
   end
+
+  def client_opportunity_matches
+    @client_opportunity_matches ||= client_opportunity_match_contacts.joins(:match)
+  end
```
eanders: This is equivalent to the relation: `client_opportunity_matches` above.  This is overriding the ActiveRecord method.  If we need different behavior, I'd suggest something like:
```suggestion
  has_many :matches, through: :client_opportunity_match_contacts, source: :match
```

app/models/contact.rb
```diff
@@ -98,4 +98,16 @@ def self.contact_type_for(input_name)
...
+
+  def open_match_count
+    client_opportunity_matches.merge(ClientOpportunityMatch.open).count
+  end
+
+  def closed_match_count
+    client_opportunity_matches.merge(ClientOpportunityMatch.closed).count
+  end
```
eanders: These are going to introduce N+1, so we should see if we can batch calculate.  I think something in the controller is probably easiest.  This should happen after the pagination.
```
@open_matches = Contact.where(id: @contacts.map(&:id)).joins(:matches).merge(ClientOpportunityMatch.open).group(:id).count
@closed_matches = Contact.where(id: @contacts.map(&:id)).joins(:matches).merge(ClientOpportunityMatch.closed).group(:id).count
```
Those would then be keyed on contact ID.

eanders (changes requested):
See inline for some thoughts.

app/models/contact.rb
gigxz: thanks, updated to use this

PR #662, opened 2022-10-19
Create new Route 9.

awisema:


app/models/notifications/nine/record_voucher_date_housing_subsidy_admin.rb
```diff
@@ -0,0 +1,23 @@
...
+module Notifications::Nine
+  class RecordVoucherDateHousingSubsidyAdmin < ::Notifications::Base
+    def self.create_for_match! match
+      match.housing_subsidy_admin_contacts.each do |contact|
+        create! match: match, recipient: contact
+      end
+    end
+
+    def decision
+      match.nine_approve_match_housing_subsidy_admin_decision
```
eanders: This doesn't feel like the right decision, but I could be wrong.  Shouldn't this be something to do with the housed date? (same with the label below)

eanders (approved):
Looks good other than one possible mis-match on the notifications.

PR #663, opened 2022-10-19
[PVT-183478468] sentry

andyzito:
https://www.pivotaltracker.com/story/show/183478468

Port Sentry implementation from warehouse. This covers catching unhandled errors in web requests, rake tasks, and background jobs.


Gemfile
```diff
@@ -39,7 +39,7 @@ gem 'memoist', require: false
 gem 'carrierwave'
 gem 'carrierwave-i18n'
 gem 'mini_magick'
-gem 'ruby-filemagic'
+# gem 'ruby-filemagic'
```
eanders: @andyzito any reason this is getting pulled out?

config/initializers/sentry.rb
```diff
@@ -0,0 +1,43 @@
...
+require_relative '../../lib/util/git'
+
+# ENV['SENTRY_DSN'] is reserved by Sentry and its use seems to prevent this initializer from being recognized.
+# Hence, we use CAS_SENTRY_DSN. Any other alternate key should also be fine.
+if ENV['CAS_SENTRY_DSN'].present?
+  Sentry.init do |config|
+    config.dsn = ENV['CAS_SENTRY_DSN']
+    config.breadcrumbs_logger = [:active_support_logger, :http_logger]
+
+    config.enabled_environments = ['production', 'staging', 'development'] # Remove development when dev is done.
```
eanders: Do we want to remove dev now?

eanders (approved):
Couple of minor questions.  I'll see if I can get this up to staging sites in a bit.

config/initializers/sentry.rb
andyzito: Not if we want to capture errors from local dev environments. I put that comment in before I realized that was a use case. I'll remove it (the comment not `development`)

Gemfile
andyzito: Cause it was preventing `bundle install` in my bare metal env and I accidentally committed it. Fixed.

PR #672, opened 2022-10-28
tagged logging port

yakloinsteak:
Also includes docker development updates to allow deployments from within containers


config/setup_logging.rb
```diff
@@ -0,0 +1,135 @@
...
+  attr_accessor :config
+
+  def initialize(config)
+    self.config = config
+  end
+
+  def run!
+    _configure_lograge
+
+    if Rails.env.development? || ENV['TEST_DEVELOPMENT_LOGGING_CONFIG'] == 'true'
```
yakloinsteak, line 13: I went for copying  from the warehouse over trying to exactly match the pre-existing logging config.

eanders (approved):

PR #673, opened 2022-10-28
Dashboard Revamp

awisema:
Just the filtering framework so far...


app/models/dashboards/overview.rb
```diff
@@ -5,6 +5,12 @@
 ###
 
 class Dashboards::Overview < Dashboards::Base
+  def initialize(filter)
+    @start_date = filter.start
+    @end_date = filter.end
```
eanders: This doesn't matter, but I wonder if delegating off to the filter would be nicer.  It would mean updating a bunch of things that take `start_date:, end_date:`, but if they took a range, you could just pass them `filter.range`, which seems way cleaner.

app/models/dashboards/overview.rb
```diff
@@ -22,7 +28,7 @@ def terminated(start_date: @start_date, end_date: @end_date)
 
   def match_results(start_date: @start_date, end_date: @end_date)
     hash = terminated(start_date: start_date, end_date: end_date).group_by(&:first)
-    hash.merge(hash) { |key, ids| ids.map { |key, id| id }}
+    hash.merge(hash) { |_, ids| ids.map { |_, id| id } }
```
eanders: I'm really confused last to what this is doing.  Is this equivalent to `ids.map(&:self)`? this is weird.

app/models/reporting/decisions.rb
```diff
@@ -5,15 +5,132 @@
 ###
 
 class Reporting::Decisions < ApplicationRecord
+  include ArelHelper
+
   belongs_to :client, foreign_key: :cas_client_id
 
-  scope :started_between, -> (start_date:, end_date:) do
+  scope :started_between, ->(start_date:, end_date:) do
```
eanders: Let's make these take a standard range object.

app/models/reporting/decisions.rb
```diff
@@ -5,15 +5,132 @@
...
+    d_2_start = at[:match_started_at]
+    d_2_end = at[:updated_at]
+    # Currently does not count as an overlap if one starts on the end of the other
+    where(d_2_end.gteq(d_1_start).or(d_2_end.eq(nil)).and(d_2_start.lteq(d_1_end)))
+  end
+
+  scope :on_route, ->(limit) do
+    route_names = Array.wrap(limit).map do |route_id|
+      MatchRoutes::Base.find(route_id).title
+    end
```
eanders: I _think_ this is the same without the N+1
```suggestion
    route_names =  MatchRoutes::Base.where(Array.wrap(limit)).pluck(:title)
```

app/models/reporting/decisions.rb
```diff
@@ -5,15 +5,132 @@
...
+    where(match_route: route_names)
+  end
+
+  scope :program_type, ->(limit) do
+    where(program_type: limit)
+  end
+
+  scope :associated_with_agency, ->(limit) do
+    # joins are hard through polymorphic associations...
+    program_names = EntityViewPermission.where(entity_type: 'Program', editable: true, agency_id: limit).
```
eanders, line 46: can you `joins(:entity)` or `preload(:entity)`?

app/models/dashboards/overview.rb
awisema: After some exploration -- I determined that this is a particularly opaque version of
```
      group_by(&:shift).
      transform_values(&:flatten)
```

app/models/reporting/decisions.rb
awisema: Unfortunately, `title` is a method overridden in the subclasses of `MatchRoute::Base` so there is no `title` to pluck. But, `find(Array.wrap(limit))` will avoid the N+1.

app/models/reporting/decisions.rb
awisema: A join results in `ActiveRecord::EagerLoadPolymorphicError: Cannot eagerly load the polymorphic association :entity` and preload doesn't actually do the load... This is where the manual join processing in the report framework comes from -- we can do it here, as we know they will all be programs, but, that seemed like an optimization pass to me.

eanders:
cas_dashboard
@awisema, I can't tag you for review since you opened this, but I believe it's ready for another pass.

awisema (commented):
Looks great! Thank you so much!

PR #676, opened 2022-11-01
Program download

eanders:


app/controllers/programs_controller.rb
```diff
@@ -69,83 +68,83 @@ def create
...
-       flash[:notice] = "New program \"<a href=\"#{edit_program_sub_program_path(@program, @sub_program)}\">#{@program.name}</a>\" created"
+      flash[:notice] = "New program \"<a href=\"#{edit_program_sub_program_path(@program, @sub_program)}\">#{@program.name}</a>\" created"
     else
-      flash[:error] = "Please review the form problems below."
+      flash[:error] = 'Please review the form problems below.'
       render :new
     end
   end
 
   private
```
awisema: I think we have been trying to move the `private` to the individual methods? RubCop doesn't like the indentation, and I think it is too hard to notice the keyword just sort of hanging out here...

app/controllers/programs_controller.rb
```diff
@@ -69,83 +68,83 @@ def create
...
-            :eligibility_requirement_notes,
-          ],
-          service_ids: [],
-          requirements_attributes: [:id, :rule_id, :positive, :variable, :_destroy]
-        )
-    end
+          :eligibility_requirement_notes,
+        ],
+        service_ids: [],
+        requirements_attributes: [:id, :rule_id, :positive, :variable, :_destroy],
```
awisema:
```suggestion
        requirements_attributes: [
          :id,
          :rule_id,
          :positive,
          :variable,
          :_destroy,
        ],
```

awisema (approved):

PR #682, opened 2022-11-15
Install Billboard JS and D3

gigxz:
* It *looks* like the deployment and build process is already set up to support yarn installation, so hopefully it will work, but I haven't tested a new image build
* Eventually we may want to move the rest of the vendor packages to be installed from the yarn registry as well. In the warehouse we did that gradually, see https://github.com/greenriver/hmis-warehouse/commits/production/vendor/assets/javascripts


gigxz:
somewhat related https://github.com/greenriver/hmis-warehouse/pull/2606

PR #695, opened 2022-12-22
Document ready rule

awisema:
Adds new rule:
image

The client must have files with all of the the specified tags to match the MUST case.

The tags are added to the list of file tags that will appear in the match files list, and are marked with a `*`:
image

### DIFF
--- a/app/models/client_opportunity_match.rb
+++ b/app/models/client_opportunity_match.rb
@@ -884,6 +884,10 @@ class ClientOpportunityMatch < ApplicationRecord
       end
   end
 
+  def associated_file_tags
+    sub_program.file_tags.pluck(:tag_id) + requirements_with_inherited.map(&:associated_file_tags).flatten.compact
+  end
+
   def self.prioritized_by_client(match_route, scope)
     match_route.match_prioritization.prioritization_for_clients(scope)
   end
--- a/app/models/client_opportunity_match.rb
+++ b/app/models/client_opportunity_match.rb
@@ -885,7 +885,16 @@ class ClientOpportunityMatch < ApplicationRecord
   end
 
   def associated_file_tags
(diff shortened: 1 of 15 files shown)


app/models/client_opportunity_match.rb
```diff
@@ -884,6 +884,10 @@ def similar_matches
       end
   end
 
+  def associated_file_tags
+    sub_program.file_tags.pluck(:tag_id) + requirements_with_inherited.map(&:associated_file_tags).flatten.compact
```
eanders: I wonder if it makes sense to expose which thing is requiring a particular tag?  As we do when we expose the rules and expose the source of the requirement. Something like:

State-issued id _(program name)_
DD-214 _(Vet Funding Source)_

app/models/rule.rb
```diff
@@ -39,12 +41,15 @@ def variable_requirement?
...
 
-  def display_for_variable value
+  def display_for_variable(_value)
     nil
   end
 
   def apply_to_match(match)
   end
+
+  def associated_file_tags(value)
```
eanders: should this return an empty array by default?
```suggestion
  def associated_file_tags(value)
    []
```

eanders (changes requested):
This looks good, just a couple of questions, and one request to add the source to the display on the match.  Thanks!

app/models/rule.rb
awisema: Doesn't really matter as the `nil`s get cleaned up at the end -- but this makes the API more consistent, and I think that is a good thing.

app/models/client_opportunity_match.rb
awisema: Like so?
image

eanders: Exactly.  Thanks!

PR #697, opened 2023-01-06
Add route specific HSA reasons for routes 8 and 9.

awisema:


app/models/match_decision_reasons/housing_subsidy_admin_decline.rb
```diff
@@ -6,6 +6,20 @@
 
 module MatchDecisionReasons
   class HousingSubsidyAdminDecline < Base
+    def self.available(include_other: true, route: nil) # rubocop:disable Lint/UnusedMethodArgument
+      other = MatchDecisionReasons::Other.first
+      # none = OpenStruct.new(name: 'None', id: nil)
+      av = active.to_a
```
eanders, line 12: is `active` a scope?  It looks like you removed them all further down.

eanders (approved):

app/models/match_decision_reasons/housing_subsidy_admin_decline.rb
awisema: Yes, `active` and `limited` are scopes on `MatchDecisionReasons::Base`. `available` uses both, and replaces `active` (or `all` in some of the code -- oops)

PR #701, opened 2023-01-10
Show available files on client page.

awisema:
Also, show file effective dates on matches.

### DIFF
--- a/app/views/clients/_files.haml
+++ b/app/views/clients/_files.haml
@@ -1,0 +1,20 @@
+.col-md-6.c-card__content-group
+  .c-card__content-group-heading
+    .c-card__content-group-icon.icon-clipboard-check
+    %h5 Files
+  - if @files.present?
+    %table.table
+      %thead
+        %tr
+          %th File
+          %th Effective Date
+      %tbody
+        - @files.each do |tag, date, id|
+          %tr
+            %td= link_to_if id.present?, tag, Config.get(:warehouse_url) + "/clients/#{@client.remote_id}/files/#{id}", target: '_blank'
+            - if date.present?
+              %td= date_format(date)
+            - else
+              %td.font-no-data None
+  - else
+    %p.font-no-data None
--- a/app/views/clients/_files.haml
+++ b/app/views/clients/_files.haml
@@ -8,12 +8,17 @@
         %tr
           %th File
           %th Effective Date
+          %th Expiration Date
       %tbody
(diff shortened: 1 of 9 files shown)


app/views/clients/_files.haml
```diff
@@ -0,0 +1,20 @@
+.col-md-6.c-card__content-group
+  .c-card__content-group-heading
+    .c-card__content-group-icon.icon-clipboard-check
+    %h5 Files
+  - if @files.present?
+    %table.table
+      %thead
+        %tr
+          %th File
+          %th Effective Date
```
eanders, line 10: Hmm, is this effective date from ClientFile in the warehouse? I believe potentially there is also an expiration date, we should show both if that's what this is since some file tags require effective date, some require expiration date, some require neither.

app/views/matches/_client_files.haml
```diff
@@ -12,8 +12,10 @@
         - if required
           *
         - if file.present?
-          - label = "#{tag} (#{file.effective_date})"
-          = link_to tag, Config.get(:warehouse_url) + "/clients/#{client.remote_id}/files/#{file.id}", target: '_blank'
+          - if file.effective_date.present?
+            - date = " (#{date_format(file.effective_date)})"
+          - label = "#{tag}#{date}"
+          = link_to label, Config.get(:warehouse_url) + "/clients/#{client.remote_id}/files/#{file.id}", target: '_blank'
```
eanders: Let's push the link calculation into the file class, this looks really brittle, but if we have it in one place, it'll be easier to maintain.

app/views/matches/_client_files.haml
```diff
@@ -12,8 +12,10 @@
         - if required
           *
         - if file.present?
-          - label = "#{tag} (#{file.effective_date})"
-          = link_to tag, Config.get(:warehouse_url) + "/clients/#{client.remote_id}/files/#{file.id}", target: '_blank'
+          - if file.effective_date.present?
+            - date = " (#{date_format(file.effective_date)})"
```
eanders: Do you think there's any point in including the date uploaded?  I'm not sure which is the most useful date, uploaded, effective, expiration.

eanders (approved):
Some possible changes inline, but otherwise I think this looks good.  Thanks!

awisema:
Ok -- Surfacing effective date and expiration date (when present):
image

image

eanders (approved):

PR #711, opened 2023-02-08
Rails 3.1.3 nominally functional

eanders:


app/controllers/system_status_controller.rb
```diff
@@ -99,20 +88,20 @@ def details
...
+      revision: Git.revision,
+      branch: Git.branch,
       hostname: `hostname`.chomp,
       cache: cache_message,
       user_count_positive: User.all.any?,
       registered_deployment_id: Rails.cache.read('registered-deployment-id'),
       environment_deployment_id: ENV['DEPLOYMENT_ID'],
       last_migration: {
-        cas: cas,
+        cas:,
```
yakloinsteak: Something's missing.

app/controllers/system_status_controller.rb
```diff
@@ -85,8 +74,8 @@ def details
       end
 
       jobs_stats[queue] = {
-        enqueued: enqueued,
-        failed: failed,
+        enqueued:,
```
yakloinsteak: Something's missing?

app/controllers/system_status_controller.rb
```diff
@@ -72,7 +61,7 @@ def details
     jobs_stats = {}
     jobs_message = 'OK'
     Delayed::Job.select('distinct queue').pluck(:queue).each do |queue|
-      scope = Delayed::Job.where(queue: queue)
+      scope = Delayed::Job.where(queue:)
```
yakloinsteak: Something's missing

yakloinsteak (changes requested):
I think there was a copy/paste error in the system status controller that needs fixing. Looks good otherwise.

app/controllers/system_status_controller.rb
eanders: You'd think.  That's ruby's new hotness in 3.1.  I'm not 100% sold on it, but rubocop is. https://blog.saeloun.com/2021/09/28/ruby-allow-value-omission-in-hash-literals.html

yakloinsteak: My professional opinion on that syntax is that it stinks. It's like `<blink>` or `goto`. I'm 0% sold on it :-)

eanders: Agreed.  Updating .rubocop.yml now

PR #717, opened 2023-03-06
Update set asides decline reasons

awisema:


app/models/concerns/match_decisions/route_four_shelter_agency_decline_reasons.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+module MatchDecisions
+  module RouteFourShelterAgencyDeclineReasons
+    extend ActiveSupport::Concern
+
+    def step_decline_reasons(_contact)
+      @step_decline_reasons ||= [
```
eanders: Why is this one `||=`?  it looks like you were using that for default reasons, but not for the others.

app/models/match_decisions/approve_match_housing_subsidy_admin.rb
```diff
@@ -113,13 +113,42 @@ def declineable_by?(contact)
...
+        'Household did not respond after initial acceptance of match',
+        'Ineligible for Housing Program',
+        'Client refused offer',
+        'Self-resolved',
+        'Falsification of documents',
+        'Additional screening criteria imposed by third parties',
+        'Health and Safety',
+        'Other',
+      ]
+      @combined_reasons ||= (@shelter_agency_contact_reasons + @hsa_contact_reasons).uniq.freeze
```
eanders: also ||= here

app/models/match_decisions/nine/nine_assign_case_contact.rb
```diff
@@ -89,6 +90,12 @@ def accessible_by? contact
         contact.in?(match.send(contact_actor_type))
     end
 
+    def step_decline_reasons(_contact)
+      [
```
eanders, line 94: Why no `@step_decline_reasons =` here, but it is in some of the others like `MatchDecisions::Nine::NineAssignManager`?

eanders (changes requested):
I think we need to standardize or document the use of instance variables for `@step_decline_reasons` and `@step_cancel_reasons`

PR #751, opened 2023-10-08
HF prioritization and cancel updates

eanders:
* Tiebreaker becomes cumulative LOT


eanders:
@awisema I'm still waiting on details around what should change for cancelation reasons, but the prioritization changes should be good to review.

awisema (approved):

PR #752, opened 2023-10-31
185607659 -- CAS match summary email

meborn:


app/mailers/match_digest_mailer.rb
```diff
@@ -0,0 +1,43 @@
...
+  def digest(contact)
+    @matches = [
+      ['Stalled', :stalled],
+      ['Recently Canceled', :canceled],
+      ['Recently Expired', :expired],
+      ['Expiring Soon', :expiring],
+      ['Active', :active],
+    ].each_with_object({}) do |group, hash|
+      hash[group[1]] = { title: group[0], matches: [] }
+    end
```
eanders: Sure seems like this would be easier to read as:
```suggestion
@matches = { 
  stalled: {
    title: 'Stalled',
    matches: [],
  },
  canceled: {
    title: 'Recently Canceled',
    matches: [],
  },
  expired: {
    title: 'Recently Expired',
    matches: [],
  },
  expiring: {
    title: 'Expiring Soon',
    matches: [],
  },
  active: {
    title: 'Active',
    matches: [],
  },
}
```

app/mailers/match_digest_mailer.rb
```diff
@@ -0,0 +1,43 @@
...
+        @matches[:expiring][:matches] << match_data.
+          merge(text: "Expiring: #{match.shelter_expiration.strftime(I18n.t('date.formats.default')) || 'unknown'}")
+      elsif match.active?
+        matched_on = match.created_at.strftime(I18n.t('date.formats.default'))
+        updated_at = match.updated_at.strftime(I18n.t('date.formats.default'))
+        @matches[:active][:matches] << match_data.
+          merge(text: "Active: #{matched_on} - #{updated_at}")
+      end
+    end
+    mail to: contact.email, subject: 'Match Digest'
```
eanders:
```suggestion
    mail to: contact.email, subject: 'Weekly CAS Match Summary'
```

app/models/client_opportunity_match.rb
```diff
@@ -377,6 +382,33 @@ def self.text_search(text)
...
+      # expiring
+      where(shelter_expiration: Date.current..Date.current + 1.week).preload(:contacts),
+      # expired
+      expired.where(shelter_expiration: 1.week.ago..).preload(:contacts),
+      # active
+      active.preload(:contacts),
+    ]
+
+    contact_ids = match_groups.map do |group|
+      group.map { |d| d.contacts.map(&:id) }.flatten
```
eanders: I think this is the same:
```suggestion
      group.flat_map { |d| d.contacts.map(&:id) }
```

eanders (approved):
Couple small tweaks.  Thanks!  This looks great.

app/models/client_opportunity_match.rb
```diff
@@ -405,7 +405,14 @@ def self.send_summary_emails
     end.flatten.uniq
 
     Contact.where(id: contact_ids).find_each do |contact|
-      MatchDigestMailer.digest(contact).deliver_now
+      if contact.user.present?
```
eanders: Let's just 
```suggestion
      return unless contact.user.present?
```
Every contact _should_ have a user, if they don't, there's no reason to send a summary, they won't be able to login.

app/controllers/accounts_controller.rb
```diff
@@ -9,16 +9,22 @@ class AccountsController < ApplicationController
...
     changed_notes = []
-    if @user.first_name != account_params[:first_name] || @user.last_name != account_params[:last_name]
-      changed_notes << "Account name was updated."
-    end
-    if @user.email_schedule != account_params[:email_schedule]
-      changed_notes << "Email schedule was updated."
+    changed_notes << 'Account name was updated.' if @user.first_name != account_params[:first_name] || @user.last_name != account_params[:last_name]
+
+    changed_notes << 'Email schedule was updated.' if @user.email_schedule != account_params[:email_schedule]
+
```
eanders, line 20: I think the following should all be behind:
`unless config.never_send_match_summary_email?`
Basically, don't let them change it if the summary emails aren't available.

PR #753, opened 2023-11-05
Me/185510516 self certification checkbox

meborn:


app/models/non_hmis_assessment.rb
```diff
@@ -110,12 +110,15 @@ def update_assessment_score!
   end
 
   def total_days_homeless_in_the_last_three_years
```
meborn, line 112: @eanders I am not sure if the logic here is correct.

app/views/non_hmis_clients/assessments/_pathways_version_three_js.haml
```diff
@@ -27,17 +27,27 @@
     var warehouse_unsheltered_selector = 'input[name*="[homeless_nights_unsheltered]"]';
     var extra_sheltered_selector = 'input[name*="[additional_homeless_nights_sheltered]"]';
     var extra_unsheltered_selector = 'input[name*="[additional_homeless_nights_unsheltered]"]';
-    var total_selector = 'input[name*="[total_days_homeless_in_the_last_three_years]"]'
+    var total_selector = 'input[name*="[total_days_homeless_in_the_last_three_years]"]';
+    var additional_days_verified_selector = 'input[name*="[self_reported_days_verified]"]'
+    var additional_days_verified_checkbox = $('input.pretty_boolean[name*="[self_reported_days_verified]"]')[0];
+
+    var calculate_total_days = function(e) {
```
meborn, line 34: @eanders Again, not sure about the logic here.

meborn:
@eanders I wasn't able to test this as much as I would have liked to. The form has a required field "Assessment Location" and I don't have any locations to choose from so I am not able to submit the form. I was able to test on the javascript side but not the backend.

eanders:
@meborn if you run this in the warehouse, do you get some locations?
```
GrdaWarehouse::Hud::Project.find_each do |project|
  GrdaWarehouse::AssessmentAnswerLookup.create(
    assessment_question: :assessment_location,
    response_code: project.id,
    response_text: project.name,
    data_source_id: project.data_source_id,
  )
end
```

app/models/non_hmis_assessment.rb
eanders: Looks right to me!  Thanks.

eanders: Hmm, it's possible `days_homeless_in_the_last_three_years` should be ignored, that may already be the sum of  `homeless_nights_sheltered` and `homeless_nights_sheltered`.

app/views/non_hmis_clients/assessments/_pathways_version_three_js.haml
eanders: Looks correct to me

app/models/non_hmis_assessment.rb
eanders: Confirmed, correct as written.  Thanks!

meborn:
Thanks @eanders that did the trick. I did a quick pass and I think this is ready but please review the logic for the days total. I am not sure I totally understood what was wanted there. I am less sure about this `app/models/non_hmis_assessment.rb#total_days_homeless_in_the_last_three_years` because there are more values to consider? The JS only looks at the 4 values in the form but this has 6 values.

eanders:
Thanks @meborn , I tracked that down. `app/models/non_hmis_assessment.rb` is handling submissions from multiple different forms so needs to take into account additional possible sources.... Oh, actually, I need to adjust so we don't accidentally cap other calculations.  I can take it from here.

PR #756, opened 2023-12-14
TC HAT updates

eanders:
This adds a few more questions to the TC HAT and calculations for required number of bedrooms, minimum occupancy, and family-ness.


awisema (approved):

eanders:
Converting this to draft until we can confirm calculations with the client. (Probably in the first week of 2024.)

DO NOT MERGE YET

PR #758, opened 2023-12-18
Mailhog removal

eanders:
This removes mailhog making it an external dependency.  It updates some documentation to point to warehouse versions (that are not yet published.)

Actual warehouse documentation is here:
Mailhog
https://github.com/greenriver/hmis-warehouse/blob/6a4100102293a21d7d0d8d31e0a6ec18728d39ce/docs/mailhog/developer-mail.md

Traefik
https://github.com/greenriver/hmis-warehouse/blob/6a4100102293a21d7d0d8d31e0a6ec18728d39ce/docs/traefik/developer-network.md

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -73,7 +73,6 @@ services:
     depends_on:
       - db
       - redis
-      - mailhog
 
   dj:
     <<: *backend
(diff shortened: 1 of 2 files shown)


dtgreiner (approved):
I updated the dc file to get traefik working in CAS and added the DIRENV notes to the developer-networking file. With these changes in, I was able to have CAS running on the traefik network.

docker/docker-compose.yml
```diff
@@ -89,9 +88,15 @@ services:
...
     command: bundle exec rails server -b 0.0.0.0
     labels:
-      - "traefik.http.routers.op.rule=Host(`open-path-cas.127.0.0.1.nip.io`)"
-      - "traefik.http.routers.op.tls=true"
-      - "traefik.http.services.op.loadbalancer.server.port=3000"
+      - traefik.enable=true
+      - traefik.http.routers.cas.entrypoints=web
+      - traefik.http.routers.cas.rule=Host(`boston-cas.dev.test`)
+      - traefik.http.services.cas_https.loadbalancer.server.port=3000
+      - traefik.http.routers.cas_https.rule=Host(`boston-cas.dev.test`)
```
eanders: @dtgreiner I _think_ we want these to reference the `FQDN` environment variable as they do in the warehouse since people have different names for the app (presumably).  Can you update that?

dtgreiner: You're right and thanks for catching that! I'm including the TRAEFIK_ENABLED environment variable as well

eanders: Thanks!  I only caught it because I couldn't get it to work locally, so I knew to check for it.  Were you able to confirm those changes worked? (after docker-compose restart)

dtgreiner: Yes, I am currently running CAS locally with this docker-config file. As long as DIRENV has been configured in the CAS environment, it should be good to go.

eanders: Ok, seems to work for me now, so I must have had something wrong.  Thanks!  Merging.

PR #760, opened 2023-12-19
Rollback Top Priority policy - 186623152

dtgreiner:

## Description

Rolls back the top priority policy

## Type of change
- [ ] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

### DIFF
--- a/app/models/client_opportunity_match.rb
+++ b/app/models/client_opportunity_match.rb
@@ -609,6 +609,14 @@ class ClientOpportunityMatch < ApplicationRecord
     Matching::RunEngineJob.perform_later
   end
 
+  def top_priority?
+    return false if closed?
+    return false unless on_or_after_first_client_step?
+    return false unless opportunity.multiple_active_matches?
+
+    self == opportunity.prioritized_active_matches.first
+  end
+
   def would_be_client_multiple_match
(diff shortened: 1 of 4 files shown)


app/models/client_opportunity_match.rb
```diff
@@ -609,14 +609,6 @@ def reset_and_destroy!
...
   end
 
-  def top_priority?
-    return false if closed?
-    return false unless on_or_after_first_client_step?
-    return false unless opportunity.multiple_active_matches?
-
-    self == opportunity.prioritized_active_matches.first
-  end
-
```
eanders: Let's leave this since it defines the logic for top_priority.  That way if we ever need to bring it back we won't have to dig too deeply.

eanders (approved):
@dtgreiner thanks for this! It looks good, but can you restore the definition of `top_priority?` I have a feeling there's a chance it'll come back and it'll be easier to restore if we keep the logic.

app/models/client_opportunity_match.rb
dtgreiner: This has been added back in.

eanders: Thanks!

PR #770, opened 2024-01-24
Determine 'Children under age 18 in household' from HAT

awisema:
Other half of https://github.com/greenriver/hmis-warehouse/pull/3941


eanders (approved):
Thank you

PR #780, opened 2024-02-16
 Build updates ported over from warehouse

yakloinsteak:
## Description

Ports over from the warehouse updates to the Dockerfiles to improve readability and build performance. Improves github action for better caching performance, vulnerability scans, and dockerfile linting.

## Type of change
- [x] Code clean-up / housekeeping


yakloinsteak:
@eanders This deploys and I can boot up things locally.

yakloinsteak:
@eanders This deploys and I can boot up things locally. I think it's ready to get moved along.

bin/rails
```diff
@@ -1,4 +1,27 @@
 #!/usr/bin/env ruby
-APP_PATH = File.expand_path('../config/application', __dir__)
-require_relative '../config/boot'
-require 'rails/commands'
```
eanders, line 4: Just noticing, but the warehouse still has bin stubs similar to this, so maybe we should apply whatever magic you did here over there as well?

docker/app/REVISION
```diff
@@ -0,0 +1 @@
+d642734d6d5be5847222e429130cf2e586c81ee1
```
eanders, line 1: Should this have been committed?
It appears to be an empty file in the warehouse.

eanders: The seems to be working fine, system_status/details is reflecting different revisions for each deployment, so probably doesn't matter.

eanders (approved):
Thanks @yakloinsteak this looks great, and I _love_ that you removed the TOFU again.

docker/app/REVISION
yakloinsteak: It just needs to exist so we could change the content to make it clearer it's just for testing

eanders: Thank you!

bin/rails
yakloinsteak: I just made https://github.com/greenriver/hmis-warehouse/pull/4057 but I won't be able to test that out.

PR #787, opened 2024-03-07
Add Match Route Prioritized Columns as selectable via the admin page

dtgreiner:

## Description

Add prioritized client columns to be selectabel via the match route admin page for each route.

## Type of change

### DIFF
--- a/app/controllers/opportunity_matches_controller.rb
+++ b/app/controllers/opportunity_matches_controller.rb
@@ -86,9 +86,7 @@ class OpportunityMatchesController < ApplicationController
         next if column_data.blank?
         next if column_data[:display_check].present? && send(column_data[:display_check]) == false
 
-        title = column_data[:title]
-        title = column_data[:alt_title] if column_data[:alt_title].present?
-        result << title
+        result << column_data[:title]
       end
     end
   end
@@ -99,7 +97,7 @@ class OpportunityMatchesController < ApplicationController
       @opportunity.match_route.prioritized_client_columns.map(&:to_sym).each do |column|
         column_data = prioritized_column_data[column]
         next if column_data.blank?
-        next if column_data[:display_check].present? && send(column_data[:display_check]) == false
+        next if column_data[:display_check].present? && ! current_user.public_send(column_data[:display_check])
 
         result << client.send(column)
       end
--- a/app/controllers/opportunity_matches_controller.rb
+++ b/app/controllers/opportunity_matches_controller.rb
@@ -75,6 +75,38 @@ class OpportunityMatchesController < ApplicationController
     }
   end
 
+  def prioritized_column_data
+    @prioritized_column_data ||= Client.prioritized_columns_data
+  end
+
+  def prioritized_column_labels
+    [].tap do |result|
+      @opportunity.match_route.prioritized_client_columns.map(&:to_sym).each do |column|
+        column_data = prioritized_column_data[column]
+        next if column_data.blank?
+        next if column_data[:display_check].present? && send(column_data[:display_check]) == false
+
+        title = column_data[:title]
+        title = column_data[:alt_title] if column_data[:alt_title].present?
+        result << title
+      end
+    end
+  end
+  helper_method :prioritized_column_labels
+
+  def prioritized_column_values(client)
+    [].tap do |result|
+      @opportunity.match_route.prioritized_client_columns.map(&:to_sym).each do |column|
+        column_data = prioritized_column_data[column]
+        next if column_data.blank?
+        next if column_data[:display_check].present? && send(column_data[:display_check]) == false
+
+        result << client.send(column)
+      end
+    end
+  end
+  helper_method :prioritized_column_values
+
   def priority_labels
     supporting_data_columns = @opportunity.match_route.match_prioritization.supporting_data_columns
     return supporting_data_columns.keys if supporting_data_columns.present?
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -594,6 +594,10 @@ class Client < ApplicationRecord
     }
   end
 
+  def calculated_is_currently_youth
+    is_currently_youth || age&.between?(18, 24) || false
+  end
+
   # Export related
   def dv_for_export
     numeric_bool_for_export(domestic_violence)
@@ -768,7 +772,6 @@ class Client < ApplicationRecord
       },
       chronic_homeless: {
         title: 'Chronically Homeless',
-        alt_title: 'Chronic',
         description: nil,
         type: 'Boolean',
       },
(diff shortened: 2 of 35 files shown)


app/controllers/opportunity_matches_controller.rb
```diff
@@ -75,6 +75,38 @@ def create_match_attributes(client_id)
...
+
+  def prioritized_column_labels
+    [].tap do |result|
+      @opportunity.match_route.prioritized_client_columns.map(&:to_sym).each do |column|
+        column_data = prioritized_column_data[column]
+        next if column_data.blank?
+        next if column_data[:display_check].present? && send(column_data[:display_check]) == false
+
+        title = column_data[:title]
+        title = column_data[:alt_title] if column_data[:alt_title].present?
```
eanders: This structure seems somewhat more complicated than it probably needs to be.  There are only 3 alt-titles.  I'll drop some notes below, but I think we can just standardize `title`

app/controllers/opportunity_matches_controller.rb
```diff
@@ -75,6 +75,38 @@ def create_match_attributes(client_id)
...
+    end
+  end
+  helper_method :prioritized_column_labels
+
+  def prioritized_column_values(client)
+    [].tap do |result|
+      @opportunity.match_route.prioritized_client_columns.map(&:to_sym).each do |column|
+        column_data = prioritized_column_data[column]
+        next if column_data.blank?
+        next if column_data[:display_check].present? && send(column_data[:display_check]) == false
```
eanders: What is this doing?

eanders: Oh, I see, it's calling things like `can_view_dob?`  I think this might be cleaner/clearer as:
```suggestion
        next if column_data[:display_check].present? && ! current_user.public_send(column_data[:display_check])
```
That will enforce that we're asking the current user (which should the current behavior, but isn't explicit) and should fail if a `nil` is returned, which I _think_ it should.

app/models/client.rb
```diff
@@ -759,6 +759,536 @@ def self.sort_options(show_vispdat: false, show_assessment: false)
...
+  def self.prioritized_columns_data
+    @prioritized_columns_data ||= {
+      veteran: {
+        title: 'Veteran',
+        description: nil,
+        type: 'Boolean',
+      },
+      chronic_homeless: {
+        title: 'Chronically Homeless',
+        alt_title: 'Chronic',
```
eanders: Pretty sure we can just use the title.
```suggestion
        title: 'Chronically Homeless',
```

app/models/client.rb
```diff
@@ -759,6 +759,536 @@ def self.sort_options(show_vispdat: false, show_assessment: false)
...
+        type: 'Integer',
+      },
+      ha_eligible: {
+        title: Translation.translate('Housing Authority Eligible'),
+        description: nil,
+        type: 'Boolean',
+      },
+      days_homeless_in_last_three_years: {
+        title: Config.using_pathways? ? 'Total # of nights' : Translation.translate('Days Homeless in Last Three Years'),
+        alt_title: 'Days homeless in the last three years',
```
eanders: I like the consistency of always using the same title if possible. so:
```suggestion
        title: Config.using_pathways? ? 'Total # of nights' : Translation.translate('Days Homeless in Last Three Years'),
```

app/models/client.rb
```diff
@@ -759,6 +759,536 @@ def self.sort_options(show_vispdat: false, show_assessment: false)
...
+        type: 'Boolean',
+      },
+      older_than_65: {
+        title: 'Manually Flagged as Over 65 ',
+        description: nil, 
+        ype: 'Boolean',
+      },
+      holds_voucher_on: {
+        title: 'Holds Mobile Voucher On',
+        alt_title: 'Date voucher was assigned',
```
eanders: I could be wrong, but I feel like it is probably fine to just lean on:
```suggestion
        title: 'Date voucher was assigned',
```

db/migrate/20240305213939_add_prioritized_client_columns_to_match_route_table.rb
```diff
@@ -0,0 +1,30 @@
...
+        # Code is in the format: "'Match Group' => ->(client) { client.match_group },"
+        code = code[code.index('{')..code.index('}')]
+        code.slice!('{ client.')
+        code.slice!('?')
+        code.slice!(' }')
+        column_list << code
+      end
+    end
+    route.constantize.update_all(prioritized_client_columns: column_list)
+  end
```
eanders: This feels a bit too clever (and probably brittle).  
There are only 4 versions of this.  Can we add a method to the schemes where we have extra columns, something like: 

in: MatchPrioritization::DaysHomelessPopulations
```
def self.supporting_column_names
  [
    :family_member,
    :veteran,
    :calculated_is_currently_youth,
    :days_homeless,
  ]
end
```

There's some complexity around the youth one, so that' would probably warrant a method on Client
```
def calculated_is_currently_youth
  is_currently_youth || age&.between?(18, 24) || false
end
```

Then, I think this migration can be something like:
```suggestion
      MatchRoutes::Base.preload(:match_prioritization).each do |route|
    route.update(prioritized_client_columns: route.match_prioritization.supporting_column_names)
  end
```

That's totally untested, so may be way off.

eanders (changes requested):
This looks great overall, thanks!  Dropped a couple of changes inline.

dtgreiner:
@eanders The requested changes have been added to the PR

app/models/match_prioritization/days_homeless_populations.rb
```diff
@@ -20,6 +20,15 @@ def self.prioritization_for_clients(scope, match_route:) # rubocop:disable Lint/
       scope.order(family, vets, is_youth, age_youth, days)
     end
 
+    def self.supporting_column_names
+      [
+        :family_member,
+        :veteran,
+        :is_currently_youth,
```
eanders: this should use the new method right?
```suggestion
        :calculated_is_currently_youth,
```

app/models/match_prioritization/days_homeless_veterans_families_youth.rb
```diff
@@ -20,6 +20,15 @@ def self.prioritization_for_clients(scope, match_route:) # rubocop:disable Lint/
       scope.order(vets, family, is_youth, age_youth, days)
     end
 
+    def self.supporting_column_names
+      [
+        :veteran,
+        :family_member,
+        :is_currently_youth,
```
eanders: Same here right?
```suggestion
        :calculated_is_currently_youth,
```

dtgreiner: yes! thanks for catching these. I've added them to the PR

eanders (approved):
Thanks @dtgreiner !

PR #816, opened 2024-06-10
Parked client download

eanders:

## Description

This:
1. Sets user, match, and reason on **Unavailable as Candidate Fors** (parking records) for existing records
2. Adds setting of the above fields where appropriate in the app to maintain them going forward
3. Adds a download feature to the Parked Clients page
4. Expose user and reason on the Parked Clients page
5. Various rubocop fixes

## Type of change

- [x] New feature (adds functionality)

- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -368,85 +368,6 @@ class Client < ApplicationRecord
     active_matches.any?
   end
 
-  def available_as_candidate_for_any_route?
-    (
-      MatchRoutes::Base.available.pluck(:type) -
-      UnavailableAsCandidateFor.where(client_id: id).distinct.pluck(:match_route_type)
-    ).any?
-  end
-
-  def make_available_in match_route:
-    route_name = MatchRoutes::Base.route_name_from_route(match_route)
-    UnavailableAsCandidateFor.where(client_id: id, match_route_type: route_name).destroy_all
-  end
-
-  def make_available_in_all_routes
-    UnavailableAsCandidateFor.where(client_id: id).destroy_all
-  end
-
-  private def default_unavailable_expiration_date
-    days = Config.get(:unavailable_for_length)
-    return nil unless days.present? && days.positive?
-
-    Date.current + days.days
-  end
-
-  def make_unavailable_in(match_route:, expires_at: default_unavailable_expiration_date)
-    route_name = MatchRoutes::Base.route_name_from_route(match_route)
-
-    un = unavailable_as_candidate_fors.where(match_route_type: route_name).first_or_create
-    un.update(expires_at: expires_at)
-  end
-
-  def make_unavailable_in_all_routes(expires_at: default_unavailable_expiration_date)
-    MatchRoutes::Base.all_routes.each do |route|
-      make_unavailable_in(match_route: route, expires_at: expires_at)
-    end
-  end
-
-  # cancel_specific must be a match object
-  def unavailable(permanent: false, contact_id: nil, cancel_all: false, cancel_specific: false, expires_at: default_unavailable_expiration_date)
-    Client.transaction do
-      update(available: false) if permanent
-
-      if cancel_all
-        # Cancel any active matches
-        client_opportunity_matches.active.each do |active_match|
-          opportunity = active_match.opportunity
-          active_match.canceled!
-          MatchEvents::Parked.create!(
-            match_id: active_match.id,
-            contact_id: contact_id,
-          )
-          opportunity.update(available_candidate: true)
-        end
-        # Delete any non-active open matches
-        client_opportunity_matches.open.each(&:delete)
-        # Prevent any new matches for this client
-        # This will re-queue the client once the date is passed
-        make_unavailable_in_all_routes(expires_at: expires_at)
-      end
-
-      if cancel_specific
-        # remove from specific match and proposed on same route
-        match = cancel_specific
-        opportunity = match.opportunity
-        route = match.match_route
-        match.canceled! # note canceled! makes the client available in the route
-        Notifications::MatchCanceled.create_for_match! match
-        MatchEvents::Parked.create!(
-          match_id: match.id,
-          contact_id: contact_id,
-        )
-        opportunity.update(available_candidate: true)
-        # Delete any non-active open matches
-        client_opportunity_matches.on_route(route).proposed.each(&:delete)
-        make_unavailable_in(match_route: route, expires_at: expires_at)
-      end
-    end
-    Matching::RunEngineJob.perform_later
-  end
-
   def remote_id
     @remote_id ||= project_client&.id_in_data_source.presence
   end
@@ -456,7 +377,7 @@ class Client < ApplicationRecord
   end
 
   def remote_data_source
-    @remote_data_source ||= DataSource.find(remote_data_source_id) rescue false # rubocop:disable Style/RescueModifier
+    @remote_data_source ||= project_client&.data_source || false
   end
 
   def remote_client_visible_to?(user)
--- a/app/models/concerns/availability.rb
+++ b/app/models/concerns/availability.rb
@@ -69,7 +69,7 @@ module Availability
       match: nil
     )
       Client.transaction do
-        update(available: false) if permanent
+        update!(available: false) if permanent
 
         if cancel_all
           # Cancel any active matches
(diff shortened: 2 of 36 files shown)


app/models/client.rb
```diff
@@ -368,85 +368,6 @@ def active_in_match?
     active_matches.any?
   end
 
-  def available_as_candidate_for_any_route?
```
eanders, line 371: This was moved into the `Availability` concern, where similar methods already existed.

app/models/concerns/availability.rb
```diff
@@ -6,24 +6,9 @@
...
-    end
-
-    def ensure_availability
-      self.class.transaction do
-        unavailable_as_candidate_fors.destroy_all
-        unavailable_routes.each do |route|
-          self.make_unavailable_in match_route: route
-        end
-      end
-    end
```
eanders, line 25: This was unused code

app/models/opportunity.rb
```diff
@@ -216,11 +216,11 @@ def confidential?
   # remove voucher
   # remove self (opportunity)
 
-  def stop_matches_and_remove_entire_history_from_existance!
+  def stop_matches_and_remove_entire_history_from_existence!
```
eanders, line 219: This had a typo, but as far as I could tell isn't used anywhere, probably a manual mechanism for data cleanup?

db/migrate/20240610185742_set_parked_client_details.rb
```diff
@@ -0,0 +1,36 @@
...
+          record.update(reason: 'Successful Match', user_id: record.versions&.first&.user_id)
+        end
+      end
+    end
+
+    # Ongoing matches
+    MatchRoutes::Base.active.each do |route|
+      matches = ClientOpportunityMatch.active.on_route(route).where(client_id: unavailable_client_ids).group_by(&:client_id)
+      UnavailableAsCandidateFor.for_route(route).where(client_id: matches.keys).find_each do |record|
+        record.update(reason: 'Active Match', user_id: record.versions&.first&.user_id, match_id: matches[record.client_id].first.id)
```
ttoomey, line 27: is matches[record.client_id] always present?

app/models/match_decisions/confirm_match_success_dnd_staff.rb
```diff
@@ -83,15 +80,13 @@ def pending
 
       def confirmed
         Notifications::MatchSuccessConfirmedDevelopmentOfficer.create_for_match! match
-        match.succeeded!
+        match.succeeded!(user: @dependencies.try(:[], :user))
```
ttoomey: Not sure but maybe it could be worth it to put this repeated code in a method on the base class?
```ruby
#   match.succeeded!(user: user)

def user
  @dependencies.try(:[], :user)
end
```

app/views/reports/parked_clients/index.haml
```diff
@@ -25,13 +29,18 @@
...
           %tbody
             - @clients.each do |client|
+              - parked = client.unavailable_as_candidate_fors.merge(UnavailableAsCandidateFor.for_route(@route))&.first
               %tr
                 %td
                   = link_to_if can_view_all_clients?, client.last_name, client_path(client)
                 %td= link_to_if can_view_all_clients?, client.first_name, client_path(client)
-                %td= client.unavailable_as_candidate_fors.merge(UnavailableAsCandidateFor.for_route(@route))&.first&.expires_at&.to_date
+                %td= parked&.expires_at&.to_date
+                %td= parked&.user&.name_with_email
```
ttoomey, line 42: preload user?

app/models/concerns/availability.rb
```diff
@@ -36,5 +21,99 @@ def active_routes
...
+      UnavailableAsCandidateFor.where(client_id: id).destroy_all
+    end
+
+    def make_unavailable_in(match_route:, user:, reason:, expires_at: default_unavailable_expiration_date, match: nil)
+      route_name = MatchRoutes::Base.route_name_from_route(match_route)
+
+      # Delete any previous unavailable_as_candidate_fors so we can track the reason for this one
+      unavailable_as_candidate_fors.where(match_route_type: route_name).destroy_all
+
+      unavailable_as_candidate_fors.where(match_route_type: route_name).create(
```
ttoomey: should these raise on fail?
```suggestion
      # Delete any previous unavailable_as_candidate_fors so we can track the reason for this one
      unavailable_as_candidate_fors.where(match_route_type: route_name).each(&:desroy!)

      unavailable_as_candidate_fors.where(match_route_type: route_name).create!(
```

app/models/concerns/availability.rb
```diff
@@ -36,5 +21,99 @@ def active_routes
...
+      permanent: false,
+      contact_id: nil,
+      cancel_all: false,
+      cancel_specific: false,
+      expires_at: default_unavailable_expiration_date,
+      user: nil,
+      match: nil
+    )
+      Client.transaction do
+        update(available: false) if permanent
```
ttoomey:
```suggestion
        update!(available: false) if permanent
```

app/models/concerns/availability.rb
```diff
@@ -36,5 +21,99 @@ def active_routes
...
+        if cancel_all
+          # Cancel any active matches
+          client_opportunity_matches.active.each do |active_match|
+            opportunity = active_match.opportunity
+            active_match.canceled!
+            MatchEvents::Parked.create!(
+              match_id: active_match.id,
+              contact_id: contact_id,
+            )
+            opportunity.update(available_candidate: true)
```
ttoomey:
```suggestion
            opportunity.update!(available_candidate: true)
```

app/models/concerns/availability.rb
```diff
@@ -36,5 +21,99 @@ def active_routes
...
+          end
+          # Delete any non-active open matches
+          client_opportunity_matches.open.each(&:delete)
```
ttoomey: these are `ClientOpportunityMatch` records? It seems like calling delete rather than destroy could leave orphaned records since delete won't trigger callbacks

app/models/concerns/availability.rb
```diff
@@ -36,5 +21,99 @@ def active_routes
...
+          match = cancel_specific
+          opportunity = match.opportunity
+          route = match.match_route
+          match.canceled! # note canceled! makes the client available in the route
+          Notifications::MatchCanceled.create_for_match! match
+          MatchEvents::Parked.create!(
+            match_id: match.id,
+            contact_id: contact_id,
+          )
+          opportunity.update(available_candidate: true)
```
ttoomey:
```suggestion
          opportunity.update!(available_candidate: true)
```

app/models/unavailable_as_candidate_for.rb
```diff
@@ -40,7 +42,22 @@ def self.describe_expiration_length
 
   def self.cleanup_expired!
     where.not(expires_at: nil).
-    where(arel_table[:expires_at].lt(Date.current)).destroy_all
+      where(arel_table[:expires_at].lt(Date.current)).
+      destroy_all
   end
 
+  def self.download_columns
```
ttoomey, line 53: This is #6133 right? Some differences here from the spec (missing Warehouse ID) but maybe that's okay?

ttoomey (approved):
The CAS models are new to me but I think I'm following along. Code looks good to me, thanks for the cleanups!

db/migrate/20240610185742_set_parked_client_details.rb
eanders: Should be.  `matches` is generated using `unavailable_client_ids` so the client ids should be a subset of the overall parked clients (those on a particular route.)  And every match and parked record will have a `client_id`

app/models/match_decisions/confirm_match_success_dnd_staff.rb
eanders: Thanks, that's cleaner for sure.

app/views/reports/parked_clients/index.haml
eanders: Not sure we can, we're loading `parked` on line 36 as an N+1, so preloading the user would already be another query.  Should be fine, these are paginated, and those queries should be quick.

app/models/unavailable_as_candidate_for.rb
eanders: Warehouse ID is actually `Remote ID` but it needs to be paired with `Remote Source` because the clients come from multiple sources.  So that should suffice.

PR #820, opened 2024-06-24
Add Match Route Eleven as Clone of Provider Only Route

dtgreiner:

## Description

Adds Match Route Eleven as a clone of the Provider Only route but with its own version of notifications, contacts, routes

## Type of change


app/models/match_decisions/eleven/eleven_confirm_hsa_accepts_client_decline_dnd_staff.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      def decline_overridden_returned
+        # Re-initialize the previous decision
+        match.eleven_hsa_accepts_client_decision.initialize_decision!
+        @decision.uninitialize_decision!(send_notifications: false)
+      end
+
+      def decline_confirmed
+        match.rejected!
+        # TODO maybe rerun the matching engine for that vacancy and client
```
eanders: I think we can remove this:
```suggestion
```

app/models/match_decisions/eleven/eleven_confirm_hsa_accepts_client_decline_dnd_staff.rb
```diff
@@ -0,0 +1,111 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+module MatchDecisions::Eleven
+  class ElevenConfirmHsaAcceptsClientDeclineDndStaff < ::MatchDecisions::Base
+    def to_partial_path
+      'match_decisions/eleven_confirm_hsa_accepts_client_decline_dnd_staff'
```
eanders: How much would it break things if we started putting the decisions into a folder?
```suggestion
      'match_decisions/eleven/confirm_hsa_accepts_client_decline_dnd_staff'
```
Seems like that would help tracking these down in the future for duplication etc.

app/models/match_decisions/eleven/eleven_hsa_acknowledges_receipt.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    def initialize_decision! send_notifications: true
+      super(send_notifications: send_notifications)
+      update status: :pending
+      send_notifications_for_step if send_notifications
+    end
+
+    def accessible_by? contact
+      contact.user_can_act_on_behalf_of_match_contacts? ||
+      contact.in?(match.housing_subsidy_admin_contacts)
```
eanders: We seem to have cleaned up `contact_actor_type` in some decisions.  Can you take a look at how we're using it in `app/models/match_decisions/approve_match_housing_subsidy_admin.rb`?  I think we want this to be:
```
contact.in?(match.send(contact_actor_type))
```
and 
```
def contact_actor_type
  nil
end
```
to be:
```
def contact_actor_type
  :housing_subsidy_admin_contacts
end
```

Looks like most (all?) of these decisions could be cleaned up that way.

app/models/match_routes/eleven.rb
```diff
@@ -0,0 +1,75 @@
...
+        'MatchDecisions::Eleven::ElevenHsaAcknowledgesReceipt' => 1,
+        'MatchDecisions::Eleven::ElevenHsaAcceptsClient' => 2,
+        'MatchDecisions::Eleven::ElevenConfirmHsaAcceptsClientDeclineDndStaff' => 2,
+      }
+    end
+
+    def required_contact_types
+      [
+        'dnd_staff',
+        'housing_subsidy_admin',
```
eanders, line 35: Should these be route specific?

app/models/match_routes/eleven.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def status_declined?(match)
+      [
+        match.eleven_hsa_accepts_client_decision&.status == 'declined' &&
+          match.eleven_confirm_hsa_accepts_client_decline_dnd_staff_decision&.status != 'decline_overridden',
+      ].any?
+    end
+
+    def contact_label_for(contact_type)
+      case contact_type
```
eanders: I wonder if this should:
```suggestion
      case contact_type&.to_sym
```

app/models/match_routes/eleven.rb
eanders: probably not, just checking.

app/views/notifications_mailer/eleven_match_initiation_for_hsa.text.haml
```diff
@@ -0,0 +1,19 @@
+Hello #{@contact.full_name},
+\
+The CAS has generated a new match recommendation.  
+\
+Please acknowledge receipt of this match and review the associated HSA contacts.
```
eanders: This should have a translation for `HSA`

app/views/notifications_mailer/hsa_decision_client.text.haml
```diff
@@ -2,7 +2,7 @@ Hello #{@contact.full_name},
 \
 = "A #{Translation.translate('Housing Subsidy Administrator')} has made a decision on a housing opportunity for which the client below applied."
 \
-Decision: #{@match.hsa_accepts_client_decision.status_label}
+Decision: #{@match.eleven_hsa_accepts_client_decision.status_label}
```
eanders: This looks like a mistake, since this is probably the original file for the provider only route

app/views/notifications_mailer/hsa_decision_hsp.text.haml
```diff
@@ -2,7 +2,7 @@ Hello #{@contact.full_name},
 \
 = "A #{Translation.translate('Housing Subsidy Administrator')} has made a decision on a housing opportunity for which the client below applied."
 \
-Decision: #{@match.hsa_accepts_client_decision.status_label}
+Decision: #{@match.eleven_hsa_accepts_client_decision.status_label}
```
eanders: This looks like a mistake, since this is probably the original file for the provider only route

app/views/notifications_mailer/hsa_decision_shelter_agency.text.haml
```diff
@@ -2,7 +2,7 @@ Hello #{@contact.full_name},
 \
 = "A #{Translation.translate('Housing Subsidy Administrator')} has made a decision on a housing opportunity for which the client below applied."
 \
-Decision: #{@match.hsa_accepts_client_decision.status_label}
+Decision: #{@match.eleven_hsa_accepts_client_decision.status_label}
```
eanders: This looks like a mistake, since this is probably the original file for the provider only route

app/views/notifications_mailer/hsa_decision_ssp.text.haml
```diff
@@ -2,7 +2,7 @@ Hello #{@contact.full_name},
 \
 = "A #{Translation.translate('Housing Subsidy Administrator')} has made a decision on a housing opportunity for which the client below applied."
 \
-Decision: #{@match.hsa_accepts_client_decision.status_label}
+Decision: #{@match.eleven_hsa_accepts_client_decision.status_label}
```
eanders: This looks like a mistake, since this is probably the original file for the provider only route

db/migrate/20240621125508_match_route_eleven_translations.rb
```diff
@@ -0,0 +1,28 @@
...
+    'SSP Eleven' => 'SSP',
+    'Shelter Agency Eleven' => 'Shelter Agency',
+    'HSA Eleven' => 'HSA',
+    'Housing Search Provider Eleven' => 'Housing Search Provider',
+    'Housing Subsidy Administrator Eleven' => 'Housing Subsidy Administrator',
+    'Stabilization Service Provider Eleven' => 'Stabilization Services Provider',
+    'Stabilization Services Provider Eleven' => 'Stabilization Services Provider',
+    'CoC Eleven' => 'DND',
+  }
+end
```
eanders: Hmm, I could be wrong, and maybe we want some default translations in there, but I _think_ you can leave this off and the app will add them via a cron job and/or if you visit something that needs it.  (It would just add the untranslated value though, so maybe something like this is useful.)

eanders (changes requested):
Thanks @dtgreiner, it generally looks really good.  
I added a couple thoughts on cleaning things up a bit, let me know if you have questions.

app/models/match_routes/eleven.rb
dtgreiner: From what I am seeing, these are not route-specific in any of the other routes. Elsewhere, we are overriding the label when displayed using reoute-specific translations.

app/views/notifications_mailer/hsa_decision_ssp.text.haml
dtgreiner: you are correct, this and the others you've noted should not be here. I'm flagging them to be revereted in the next commit.

app/models/match_decisions/eleven/eleven_confirm_hsa_accepts_client_decline_dnd_staff.rb
dtgreiner: From an initial test, it doen't look like it is breaking anything, but I'll do a bit more to get some more thorough results before including in in the next commit.

db/migrate/20240621125508_match_route_eleven_translations.rb
dtgreiner: From my understanding the app will automatically add these transalations. I'll remove this file and we can add it back in if needed. I saw a previous match route setting the default translations via script so included it here based on that knowledge.

eanders (approved):
Thanks @dtgreiner, looks great!

PR #823, opened 2024-07-01
filter matches by contact

dtgreiner:

## Description

When the contact and/or. the contact type filter are selected for the list of opportunities, use the filter to determine which associated matches are shown. 

If just the user -> only include matches where the user is a contact 
If just the type -> include any match that has a person filling that ccontact type
If both -> only show matches where the selected user is in the selected contact type

## Type of change


eanders (approved):
I think this looks good to me.  Oddly, my local dev seems to decide the sort order is different based on the filter, but I think only because I'm removing the match that was touched most recently.

PR #824, opened 2024-07-09
Allow automatic parking of configed routes for active and successful matches

dtgreiner:

## Description

Add routes to park on active and successful matches to the match route configurations. On a match being active or successful, park the routes specified in the config for the corresponding event. Similarly, if an active route is no longer active (canceled/rejected/poached) unpark the routes that were parked due to the match becoming active.

## Type of change

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -456,6 +456,10 @@ class Client < ApplicationRecord
     available
   end
 
+  def unavailable_fors_include_active_match
+    unavailable_as_candidate_fors.detect { |uf| uf.reason == UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT }.present?
+  end
+
   def available_text
     if is_available_for_matching?
       if available_as_candidate_for_any_route?
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -456,7 +456,7 @@ class Client < ApplicationRecord
     available
   end
 
-  def unavailable_fors_include_active_match
+  def unavailable_fors_include_active_match?
     unavailable_as_candidate_fors.detect { |uf| uf.reason == UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT }.present?
   end
 
--- a/app/models/client_opportunity_match.rb
+++ b/app/models/client_opportunity_match.rb
@@ -657,7 +657,7 @@ class ClientOpportunityMatch < ApplicationRecord
 
   def reopen!(contact, user: nil)
     self.class.transaction do
-      client.make_unavailable_in(match_route: match_route, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT)
+      # Park client on any other routes where this route is set to block matching when the client is involved in this route
       match_route.routes_parked_on_active_match.reject(&:empty?).each do |park_route|
         client.make_unavailable_in(match_route: park_route.constantize, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT)
       end
@@ -675,7 +675,7 @@ class ClientOpportunityMatch < ApplicationRecord
   def activate!(touch_referral_event: true, user: nil)
     self.class.transaction do
       update!(active: true)
-      client.make_unavailable_in(match_route: match_route, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT) if match_route.should_prevent_multiple_matches_per_client
+      # Park client on any other routes where this route is set to block matching when the client is involved in this route
       match_route.routes_parked_on_active_match.reject(&:empty?).each do |park_route|
         client.make_unavailable_in(match_route: park_route.constantize, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT)
       end
@@ -701,7 +701,6 @@ class ClientOpportunityMatch < ApplicationRecord
   def rejected!(touch_referral_event: true)
     self.class.transaction do
       update! active: false, closed: true, closed_reason: 'rejected'
-      client.make_available_in(match_route: match_route)
       unpark_routes_parked_for_active_match
       opportunity.update! available_candidate: !opportunity.active_matches.exists?
       RejectedMatch.create! client_id: client.id, opportunity_id: opportunity.id
@@ -717,7 +716,6 @@ class ClientOpportunityMatch < ApplicationRecord
   def canceled!(touch_referral_event: true)
     self.class.transaction do
       update! active: false, closed: true, closed_reason: 'canceled'
-      client.make_available_in(match_route: match_route)
       unpark_routes_parked_for_active_match
       opportunity.update! available_candidate: !opportunity.active_matches.exists?
       RejectedMatch.create! client_id: client.id, opportunity_id: opportunity.id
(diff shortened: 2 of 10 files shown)


dtgreiner:
I missed updating the language requirements below - will get a new commit in here soon with that update.

-Language on CAS Client page should be updated to indicate if a client is unavailable because they are currently in a match or if they are in a successful match.
-IF client is in a match (but NOT completed) should read: _Clients are blocked from matching while actively in a resource match.
{client name] will not match any vacancies on the following routes:
-IF client is in a successful match, language can stay as it currently is

app/models/client.rb
```diff
@@ -456,6 +456,10 @@ def is_available_for_matching? # rubocop:disable Naming/PredicateName
     available
   end
 
+  def unavailable_fors_include_active_match
```
eanders: maybe:
```suggestion
  def unavailable_fors_include_active_match?
```
Just to indicate it returns a boolean?

app/models/client_opportunity_match.rb
```diff
@@ -658,6 +658,9 @@ def restrictions_on_reopening
   def reopen!(contact, user: nil)
     self.class.transaction do
       client.make_unavailable_in(match_route: match_route, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT)
+      match_route.routes_parked_on_active_match.reject(&:empty?).each do |park_route|
```
eanders, line 661:
```suggestion
      # Park client on any other routes where this route is set to block matching when the client is involved in this route
      match_route.routes_parked_on_active_match.reject(&:empty?).each do |park_route|
```

app/models/client_opportunity_match.rb
```diff
@@ -673,6 +676,9 @@ def activate!(touch_referral_event: true, user: nil)
     self.class.transaction do
       update!(active: true)
       client.make_unavailable_in(match_route: match_route, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT) if match_route.should_prevent_multiple_matches_per_client
+      match_route.routes_parked_on_active_match.reject(&:empty?).each do |park_route|
```
eanders, line 679:
```suggestion
      # Park client on any other routes where this route is set to block matching when the client is involved in this route
      match_route.routes_parked_on_active_match.reject(&:empty?).each do |park_route|
```

app/models/client_opportunity_match.rb
```diff
@@ -673,6 +676,9 @@ def activate!(touch_referral_event: true, user: nil)
...
+        client.make_unavailable_in(match_route: park_route.constantize, expires_at: nil, user: user, match: self, reason: UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT)
```
eanders, line 680: Pretty sure this shouldn't be `UnavailableAsCandidateFor::ACTIVE_MATCH_TEXT`  right?

app/models/client_opportunity_match.rb
```diff
@@ -767,6 +776,10 @@ def succeeded!(touch_referral_event: true, user: nil)
         client.make_unavailable_in(match_route: route, user: user, match: self, reason: UnavailableAsCandidateFor::SUCCESSFUL_MATCH_TEXT)
       end
 
+      match_route.routes_parked_on_successful_match.reject(&:empty?).each do |park_route|
```
eanders, line 773:
```suggestion
      # Park client on any other routes where this route is set to block matching when the client is successful on this route
      match_route.routes_parked_on_successful_match.reject(&:empty?).each do |park_route|
```

eanders (approved):
This looks great, thanks @dtgreiner !

I dropped a few thoughts, mostly more code comments.
Do you think it would be hard to add some basic tests?  I think there should be a couple, basically, set the config to park on another route while active, and a different one when successful, then confirm the unavailable fors get generated/removed as appropriate.

eanders:
Oh, and I think we need a migration that sets the defaults to the current behavior.  I _think_ that is just to park on the same route as the match, but we should double check that.

dtgreiner:
> Oh, and I think we need a migration that sets the defaults to the current behavior. I _think_ that is just to park on the same route as the match, but we should double check that.

I left in the code that was automatically parking the client on the current route on Active and Successful matches, but I agree it feels cleaner to include it in the new configs instead. I'll pull those out of the code and add them in as defaults for the new configs in the migration. 

The only other place we are parking routes in this file is in the `succeeded!` method when the `should_cancel_other_matches` flag is set, we are parking all other routes. I'll plan on leaving that functionality as is.

eanders:
> > Oh, and I think we need a migration that sets the defaults to the current behavior. I _think_ that is just to park on the same route as the match, but we should double check that.
> 
> I left in the code that was automatically parking the client on the current route on Active and Successful matches, but I agree it feels cleaner to include it in the new configs instead. I'll pull those out of the code and add them in as defaults for the new configs in the migration.

That sounds good ^^^.

> The only other place we are parking routes in this file is in the `succeeded!` method when the `should_cancel_other_matches` flag is set, we are parking all other routes. I'll plan on leaving that functionality as is.

Since we have configuration for park on routes on success now, it seems reasonable to pull that from code and use the configuration.  I may be mis-reading that though.  `should_cancel_other_matches` should still happen, that's about canceling matches not parking the client, but if we park them on the current route on success, I think we should do that based on config instead of in code.

Thanks!

dtgreiner:
> > > Oh, and I think we need a migration that sets the defaults to the current behavior. I _think_ that is just to park on the same route as the match, but we should double check that.
> > 
> > 
> > I left in the code that was automatically parking the client on the current route on Active and Successful matches, but I agree it feels cleaner to include it in the new configs instead. I'll pull those out of the code and add them in as defaults for the new configs in the migration.
> 
> That sounds good ^^^.
> 
> > The only other place we are parking routes in this file is in the `succeeded!` method when the `should_cancel_other_matches` flag is set, we are parking all other routes. I'll plan on leaving that functionality as is.
> 
> Since we have configuration for park on routes on success now, it seems reasonable to pull that from code and use the configuration. I may be mis-reading that though. `should_cancel_other_matches` should still happen, that's about canceling matches not parking the client, but if we park them on the current route on success, I think we should do that based on config instead of in code.
> 
> Thanks!

Right now, if `should_cancel_other_matches` is true, the match is canceling all other matches and then parking all other routes, regardless of whether or not there was a match on the route. If it is false, its just canceling the current route. I can write the migration to check for this flag and include all routes in the default if it is set or only the current route if it is not set.

eanders:
> Right now, if `should_cancel_other_matches` is true, the match is canceling all other matches and then parking all other routes, regardless of whether or not there was a match on the route. If it is false, its just canceling the current route. I can write the migration to check for this flag and include all routes in the default if it is set or only the current route if it is not set.

That sounds like the correct behavior for the migration.  The cancelation behavior should remain, but only park on routes specified.  Thanks!

app/models/client.rb
dtgreiner: This has been included in the lates commit

app/models/client_opportunity_match.rb
dtgreiner: This has been included in the lates commit

app/models/client_opportunity_match.rb
dtgreiner: This has been included in the lates commit

app/models/client_opportunity_match.rb
dtgreiner: This has been included in the lates commit

dtgreiner:
@eanders I've added the defaults to the migration and removed them from the code. The only other flag that was affecting this was `should_prevent_multiple_matches_per_client`. I have the migration defaults for the active route parking to only include the current route when this flag is set. This is how the code was working in the `activate!` method

dtgreiner:
@eanders the test should all be successful now

eanders (approved):
Thanks @dtgreiner !

PR #852, opened 2024-08-27
Add Program Filter to Dashboard Report

dtgreiner:

## Description

Adds a new filter to the CAS Dashboard Overview report to allow filtering by Program

## Type of change

### DIFF
--- a/app/models/reporting/decisions.rb
+++ b/app/models/reporting/decisions.rb
@@ -43,7 +43,7 @@ class Reporting::Decisions < ApplicationRecord
   end
 
   scope :program, ->(limit) do
-    where(program_name: Program.find(limit).pluck(:name))
+    where(program_name: Program.where(id: limit).select(:name))
   end
 
   scope :associated_with_agency, ->(limit) do
--- a/app/models/reporting/decisions.rb
+++ b/app/models/reporting/decisions.rb
@@ -42,6 +42,10 @@ class Reporting::Decisions < ApplicationRecord
     where(program_type: limit)
   end
 
+  scope :program, ->(limit) do
+    where(program_name: Program.find(limit).pluck(:name))
+  end
+
   scope :associated_with_agency, ->(limit) do
     # joins are hard through polymorphic associations...
     program_names = EntityViewPermission.where(entity_type: 'Program', editable: true, agency_id: limit).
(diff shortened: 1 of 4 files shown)


app/models/reporting/decisions.rb
```diff
@@ -42,6 +42,10 @@ class Reporting::Decisions < ApplicationRecord
     where(program_type: limit)
   end
 
+  scope :program, ->(limit) do
+    where(program_name: Program.find(limit).pluck(:name))
```
eanders: Hmm, so, I'm not super excited about matching on program name, though I suppose you can't tell the difference between two identically named programs in the drop down. Why are you plucking here?  Why not?
```suggestion
    where(program_name: Program.where(id: limit).select(:name))
```

app/models/reporting/filter_base.rb
```diff
@@ -137,6 +142,10 @@ def match_route_options_for_select
       end.to_h
     end
 
+    def program_options_for_select
+      Program.order(name: :asc).pluck(:name, :id).to_h
```
eanders, line 146: calling `.to_h` here will throw out duplicates based on name.  Since the match is by name at the end of the day, (above) maybe that's ok, but it's a bit odd.  Let's leave it for now.

app/views/reports/dashboards/_report_filter.haml
```diff
@@ -15,6 +15,8 @@
         = f.input :match_route, as: :select_2, collection: @filter.match_route_options_for_select,  selected: @filter.match_routes&.first, include_blank: false
     .col-4
       = f.input :program_types, as: :select_2, collection: @filter.program_type_options_for_select, input_html: { multiple: true, placeholder: 'Any Program Type', data: { placeholder: 'Any Program Type' } }
+    .col-4
```
eanders: Can you update these all to `.col-3` so we end up with 4 items on the row to match the client-level filters below?
Screenshot 2024-08-27 at 10 26 22 AM

app/models/reporting/decisions.rb
dtgreiner: Reporting::Decisions is storing the program name and not the id. I agree filtering by id would be much better, but I didn't see a great way to get there with the current table setup. 
I can switch this over to a select in the next commmit.

app/views/reports/dashboards/_report_filter.haml
dtgreiner: Will include this in the neext commit

app/models/reporting/filter_base.rb
dtgreiner: I cloned this one from agency_options_for_select below, but happy to switch it up if that makes more sense.

eanders (approved):

PR #859, opened 2024-09-16
Add 'Primary Option' for each contact type on Matches

dtgreiner:

## Description

Adds the option to set a primary contact for each contact type in a match. This works in the modal and non-modal version of the edit match contacts and displays the primary contact in the contacts partial displayed on the match page.

The first match added as a contact for each type is automatically designated as the primary contact.

A button to filter just primary contacts is also included in the Matches by Contacts dashboard report.

## Type of change

### DIFF
--- a/app/controllers/match_contacts_controller.rb
+++ b/app/controllers/match_contacts_controller.rb
@@ -51,23 +51,21 @@ class MatchContactsController < ApplicationController
   private
 
   def set_contact_order(update_contact, order_number = 1)
-    join_contact = match.client_opportunity_match_contacts.find(update_contact)
-    contact_type = nil
-    match_contact_methods.each do |type, _|
-      contact_type = type if join_contact.send(type)
-    end
+    ClientOpportunityMatchContact.transaction do
+      join_contact = match.client_opportunity_match_contacts.find(update_contact)
+      contact_type = nil
+      match_contact_methods.each do |type, _|
+        contact_type = type if join_contact.send(type)
+      end
 
-    # We currently only want one contact per order number, so strip order status for any other contacts in the group
-    a_t = ClientOpportunityMatchContact.arel_table
-    contacts = @match.client_opportunity_match_contacts.where(a_t[contact_type].eq(true)).where(contact_order: order_number)
-    contacts.each do |contact|
-      contact.contact_order = nil
-      contact.save
-    end
+      # We currently only want one contact per order number, so strip order status for any other contacts in the group
+      a_t = ClientOpportunityMatchContact.arel_table
+      @match.client_opportunity_match_contacts.where(a_t[contact_type].eq(true)).where(contact_order: order_number).update_all(contact_order: nil)
 
-    # Set the selected contact as ordered contact
-    join_contact.contact_order = order_number
-    join_contact.save
+      # Set the selected contact as ordered contact
+      join_contact.contact_order = order_number
+      join_contact.save!
+    end
   end
 
   def match_contact_methods
--- a/app/controllers/match_contacts_controller.rb
+++ b/app/controllers/match_contacts_controller.rb
@@ -22,72 +22,116 @@ class MatchContactsController < ApplicationController
   def update
     update_params = match_contacts_params
     saved = @match_contacts.update(update_params)
-    unless request.xhr?
-      if saved
-        flash[:notice] = "Match Contacts updated"
-        redirect_to match_path(@match)
-      else
-        raise @match_contacts.errors.full_messages.inspect
-        flash[:error] = "Please review the form problems below."
-        redirect_to :edit
-      end
+    return if request.xhr?
+
+    if saved
+      flash[:notice] = 'Match Contacts updated'
+      redirect_to match_path(@match)
+    else
+      raise @match_contacts.errors.full_messages.inspect
+      flash[:error] = 'Please review the form problems below.' # rubocop:disable Lint/UnreachableCode
+      redirect_to :edit
     end
+  end
+
+  def reorder
+    return unless params[:join_contact]
+
+    order_number = 1 # Default to setting the contact as primary
+    order_number = params[:order_number].to_i if params[:order_number]
+    set_contact_order(params[:join_contact], order_number)
+    redirect_to match_path(@match)
   end
 
   private
 
-    def match_scope
-      ClientOpportunityMatch.all
+  def set_contact_order(update_contact, order_number = 1)
+    join_contact = match.client_opportunity_match_contacts.find(update_contact)
+    contact_type = nil
+    match_contact_methods.each do |type, _|
+      contact_type = type if join_contact.send(type)
     end
 
-    def set_match
-      @match = match_scope.find params[:match_id]
+    # We currently only want one contact per order number, so strip order status for any other contacts in the group
+    a_t = ClientOpportunityMatchContact.arel_table
+    contacts = @match.client_opportunity_match_contacts.where(a_t[contact_type].eq(true)).where(contact_order: order_number)
+    contacts.each do |contact|
+      contact.contact_order = nil
+      contact.save
     end
 
-    def set_current_contact
-      @current_contact = current_contact
-    end
+    # Set the selected contact as ordered contact
+    join_contact.contact_order = order_number
+    join_contact.save
+  end
 
-    def set_match_contacts
-      @match_contacts = @match.match_contacts
-    end
+  def match_contact_methods
+    @match_contact_methods ||= {
+      shelter_agency: :shelter_agency_contacts,
+      client: :client_contacts,
+      dnd_staff: :dnd_staff_contacts,
+      housing_subsidy_admin: :housing_subsidy_admin_contacts,
+      ssp: :ssp_contacts,
+      hsp: :hsp_contacts,
+      do: :do_contacts,
+    }
+  end
 
-    def match_contacts_params
-      base_params = params[:match_contacts] || ActionController::Parameters.new
-      base_params.permit(
-        shelter_agency_contact_ids: [],
-        housing_subsidy_admin_contact_ids: [],
-        dnd_staff_contact_ids: [],
-        client_contact_ids: [],
-        ssp_contact_ids: [],
-        hsp_contact_ids: [],
-        do_contact_ids: [],
-      ).tap do |result|
-        if current_contact.user_can_edit_match_contacts?
-          result[:shelter_agency_contact_ids] ||= []
-          result[:client_contact_ids] ||= []
-          result[:dnd_staff_contact_ids] ||= []
-          result[:housing_subsidy_admin_contact_ids] ||= []
-          result[:ssp_contact_ids] ||= []
-          result[:hsp_contact_ids] ||= []
-          result[:do_contact_ids] ||= []
-        elsif hsa_can_edit_contacts?
-          # only allow editing of the hsa contacts
-          result[:shelter_agency_contact_ids] = @match.shelter_agency_contact_ids
-          result[:client_contact_ids] = @match.client_contact_ids
-          result[:dnd_staff_contact_ids] = @match.dnd_staff_contact_ids
-          result[:housing_subsidy_admin_contact_ids] ||= []
-          result[:ssp_contact_ids] = @match.ssp_contact_ids
-          result[:hsp_contact_ids] = @match.hsp_contact_ids
-          result[:do_contact_ids] = @match.do_contact_ids
-          # always add self
-          result[:housing_subsidy_admin_contact_ids] << current_contact.id unless result[:housing_subsidy_admin_contact_ids].map(&:to_i).include? current_contact.id
-        end
+  def match_contacts_method_for input_name
+    match_contact_methods[input_name] || input_name
+  end
+
+  def match_scope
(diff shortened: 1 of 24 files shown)


app/controllers/match_contacts_controller.rb
```diff
@@ -22,72 +22,121 @@ def edit
...
+    contact_type = nil
+    match_contact_methods.each do |type, _|
+      contact_type = type if join_contact.send(type)
     end
 
-    def set_match
-      @match = match_scope.find params[:match_id]
+    # We currently only want one contact per order number, so strip order status for any other contacts in the group
+    a_t = ClientOpportunityMatchContact.arel_table
+    contacts = @match.client_opportunity_match_contacts.where(a_t[contact_type].eq(true)).where(contact_order: order_number)
```
eanders: Do you need to drop into Arel here?
This looks like it could be:
```
contacts = @match.client_opportunity_match_contacts.where(contact_type: true, contact_order: order_number)
```

eanders: This whole section should probably be wrapped in a transaction so that if some part of it fails, it will be atomic.
```
ClientOpportunityMatchContact.transaction do
...
end
```

Also, rather than pull back the contacts and then save them each one at a time:
```
@match.client_opportunity_match_contacts.where(a_t[contact_type].eq(true)).where(contact_order: order_number).update_all(contact_order: nil)
```
should accomplish that in one query, unless there are callbacks that need to be run.

app/controllers/match_contacts_controller.rb
```diff
@@ -22,72 +22,121 @@ def edit
...
+      contact.contact_order = nil
+      contact.save
     end
 
-    def set_current_contact
-      @current_contact = current_contact
-    end
+    # Set the selected contact as ordered contact
+    join_contact.contact_order = order_number
+    join_contact.save
```
eanders: Probably worth making this 
```suggestion
    join_contact.save!
```
to make sure it errors if it is unable to set the order.

app/controllers/match_contacts_controller.rb
```diff
@@ -22,72 +22,121 @@ def edit
...
+  def match_contact_methods
+    @match_contact_methods ||= {
+      shelter_agency: :shelter_agency_contacts,
+      client: :client_contacts,
+      dnd_staff: :dnd_staff_contacts,
+      housing_subsidy_admin: :housing_subsidy_admin_contacts,
+      ssp: :ssp_contacts,
+      hsp: :hsp_contacts,
+      do: :do_contacts,
+    }
```
eanders, line 79: Can we pull these from somewhere in the `ClientOpportunityMatch` or `ClientOpportunityMatchContact` models?  It makes me a bit nervous given that these have evolved over time.  I'd like to centralize that near where we define these types if not re-use some existing method.

app/models/match_contacts.rb
```diff
@@ -118,6 +148,19 @@ def input_names
...
+  def join_contacts_method_for input_name
+    @join_contacts_method_for ||= {
+      shelter_agency_contacts: :shelter_agency_join_contacts,
+      client_contacts: :client_join_contacts,
+      dnd_staff_contacts: :dnd_staff_join_contacts,
+      housing_subsidy_admin_contacts: :housing_subsidy_admin_join_contacts,
+      ssp_contacts: :ssp_join_contacts,
+      hsp_contacts: :hsp_join_contacts,
+      do_contacts: :do_join_contacts,
+    }
```
eanders: Same with this list, can we centralize it?

app/models/reporting/decisions.rb
```diff
@@ -160,6 +160,14 @@ class Reporting::Decisions < ApplicationRecord
       merge(contacts)
   end
 
+  scope :contacts_in_order, ->(contact_array, order) do
+    contacts = ClientOpportunityMatchContact.where(contact_order: order)
+    contacts.where(contact_id: contact_array) unless contact_array.blank?
+
+    joins(match: :contacts).
+      merge(contacts)
+  end
```
eanders: Should this return an ordered scope?  The name is a bit odd.  I think I get what's going on, but maybe `contacts_with_order_value` would be more clear.

app/controllers/match_contacts_controller.rb
dtgreiner: The update_all and transaction definitely make sense. I'll ge them included. 

As for the arel, I was running into an issue getting the AR query to work with the column name being variable (contact_type) so used arel. Happy to switch that up if there's a better way to do it.

app/controllers/match_contacts_controller.rb
dtgreiner: will include this in the next commit

app/models/reporting/decisions.rb
dtgreiner: makes sense to me, I'll update the name to make it more clear.

app/models/match_contacts.rb
dtgreiner: If we pull this list to a more centralized location (e.g. `ClientOpportunityMatchContact`), do we want to do the same with the `available_contacts_method_for` & `selected_contacts_method_for` lists that were already in this file too?

dtgreiner: Looking at it more, leaving those other two lists may make sense. They have overrides in the `client_contacts`, `program_contacts`, & `sub_program_contacts` files

eanders: If we're generating the same hash in multiple places, let's figure out how to centralize it.  If they are different only on the value side, let's use the centralized method, but replace the values as appropriate.  (Let me know if that's not clear.)  If they are different on the key side, it probably makes sense to keep them separate.  Just trying to keep from generating a bunch of these hashes that need to be kept in sync in the future.

app/views/match_contacts/_match_contacts_list.haml
```diff
@@ -9,12 +9,17 @@
       - Translation.translate('Stabilization Service Providers')
       - Translation.translate('Housing Search Providers')
       .c-detail-group__title--secondary= Translation.translate(title)
-      .detail-box--value
-        - match.send(contact_type).each do |contact|
+      .detail-box--value 
+        - contact_type_column = contact_type.to_s.sub('_contacts','').to_sym #TODO: This feels too hacky
```
eanders: Can we move this into the `ClientOpportunityMatchContact` class?  It could have a method:
```
ClientOpportunityMatchContact.column_for(contact_type)
```

Ideally that would lookup the column in a has keyed on type.

app/views/match_contacts/_match_contacts_list.haml
```diff
@@ -9,12 +9,17 @@
...
+        - all_contacts = match.client_opportunity_match_contacts.where(ClientOpportunityMatchContact.arel_table[contact_type_column.to_sym].eq(true)).index_by(&:contact_id)
```
eanders: Alternatively (or maybe in addition), there could be a scope on `ClientOpportunityMatchContact` that takes the `contact_type` and returns where the column is true.  This could then be:
```suggestion
        - all_contacts = match.client_opportunity_match_contacts.for_contact_type(contact_type).index_by(&:contact_id)
```

eanders:
Thanks @dtgreiner this looks good overall.  Dropped a few cleanup suggestions inline, but also wanted to call out a couple of UI changes we should make.

1. ~~There isn't any indication (that I could find) that says what the checkbox or litle person icon are for.~~  Oh, I just looked again and the tooltips are showing up now.  That might be sufficient.

Screenshot 2024-09-16 at 9 46 12 AM


Screenshot 2024-09-16 at 12 47 58 PM

Screenshot 2024-09-16 at 12 50 27 PM

2. When there is only one contact in a given category, can we hide the checkbox and head icon?  I don't think we need to indicate that they are the primary if there isn't anyone else.

Screenshot 2024-09-16 at 12 51 28 PM

Screenshot 2024-09-16 at 12 51 36 PM

3.  Turns out the issue with the tooltips seems to be related to ajax loading of the Contacts section on the main match page.  It isn't re-enabling tooltips after the save from the modal.

dtgreiner:
@eanders I have the requested updates for this added to the PR 


The only thing I've left out for now is any changes to the existing `available_contacts_method_for` & `selected_contacts_method_for` lists.

The `program_contacts`, `sub_program_contacts`, & `match_contacts` lists have the same keys & values, but the `client_contacts` has one differing key from the rest.


**Client Contacts List**
```
@available_contacts_methods ||= {
  shelter_agency_contacts: :available_shelter_agency_contacts,
  regular_contacts: :available_regular_contacts,
  dnd_staff_contacts: :available_dnd_staff_contacts,
  housing_subsidy_admin_contacts: :available_housing_subsidy_admin_contacts,
  ssp_contacts: :available_ssp_contacts,
  hsp_contacts: :available_hsp_contacts,
  do_contacts: :available_do_contacts,
}
```

**Match, Program, SubProgam Lists**
```
@available_contacts_methods ||= {
  shelter_agency_contacts: :available_shelter_agency_contacts,
  client_contacts: :available_client_contacts,
  dnd_staff_contacts: :available_dnd_staff_contacts,
  housing_subsidy_admin_contacts: :available_housing_subsidy_admin_contacts,
  ssp_contacts: :available_ssp_contacts,
  hsp_contacts: :available_hsp_contacts,
  do_contacts: :available_do_contacts,
}
```

I can definitely clean these up a bit, but wanted to get the rest of the update requests in first.

app/controllers/match_contacts_controller.rb
```diff
@@ -22,72 +22,119 @@ def edit
...
+      match_contact_methods.each do |type, _|
+        contact_type = type if join_contact.send(type)
+      end
 
-    def set_match
-      @match = match_scope.find params[:match_id]
-    end
+      # We currently only want one contact per order number, so strip order status for any other contacts in the group
+      a_t = ClientOpportunityMatchContact.arel_table
+      @match.client_opportunity_match_contacts.where(a_t[contact_type].eq(true)).where(contact_order: order_number).update_all(contact_order: nil)
```
eanders: Pretty sure this can be cleaned up to:
```suggestion
      @match.client_opportunity_match_contacts.where(contact_type => true, contact_order: order_number).update_all(contact_order: nil)
```
Am I missing something there?

app/models/client_opportunity_match_contact.rb
```diff
@@ -16,14 +16,46 @@ class ClientOpportunityMatchContact < ApplicationRecord
 
   acts_as_paranoid
 
+  scope :for_contact_type, ->(contact_type) do
+    contact_type_column = ClientOpportunityMatchContact.column_for(contact_type)
+    where(ClientOpportunityMatchContact.arel_table[contact_type_column.to_sym].eq(true))
```
eanders: Same here:
```suggestion
    where(contact_type_column => true)
```

app/models/client_opportunity_match_contact.rb
```diff
@@ -16,14 +16,46 @@ class ClientOpportunityMatchContact < ApplicationRecord
...
+    contact_type_columns[contact_type] || contact_type
+  end
+
+  def self.join_contact_methods
+    contact_types.map { |t| ["#{t}_contacts".to_sym, "#{t}_join_contacts".to_sym] }.to_h
+  end
+
+  def self.join_method_for(contact_type)
+    join_contact_methods[contact_type] || contact_type
+  end
```
eanders, line 60: This feels good.  Thanks.

app/controllers/match_contacts_controller.rb
dtgreiner: you're right. I must have been getting the syntax wrong. I was trying to do this without arel and running into issues with the variable contact type. I'll ge this and the `client_opportunity_match_contact.rb` one switched up

app/models/client_opportunity_match_contact.rb
dtgreiner: Do we want to clean up the client, program, and subprogram lists too, or do you think its better to leave them as is for now?

app/controllers/match_contacts_controller.rb
dtgreiner: @eanders This has been added to the PR

app/models/client_opportunity_match_contact.rb
dtgreiner: @eanders This has been added to the PR

app/models/client_opportunity_match_contact.rb
eanders: If they follow a similar pattern and we can be fairly certain we're not breaking anything, then yes, please!

dtgreiner: I just pushed a new commit with the list updates for the additional contact files

eanders:
@dtgreiner one, hopefully final, request.  Can you add a migration to set anyone who is the only contact of the type to be the primary, for any historic or active matches?  If there are 2 or more contacts of a type, just leave it unset for now.  I think we need this for the reporting to make sense.

dtgreiner:
> @dtgreiner one, hopefully final, request. Can you add a migration to set anyone who is the only contact of the type to be the primary, for any historic or active matches? If there are 2 or more contacts of a type, just leave it unset for now. I think we need this for the reporting to make sense.

@eanders This should be all set to go.

eanders (approved):
Thanks @dtgreiner !

PR #894, opened 2024-12-23
Match progress report - support for duplicates program names

eanders:
## _Merging this PR_

## Description
* Fixes an issue where programs and sub-programs with exactly the same names were prevented from showing up in the match progress report

## Type of change
* Bug fix


app/controllers/reports/match_progress_controller.rb
```diff
@@ -19,7 +19,7 @@ def index
       respond_to do |format|
         format.html {}
         format.xlsx do
-          @included_sub_programs = sub_program_list.invert.slice(*report_params[:sub_programs])
+          @included_sub_programs = sub_program_list.select { |sp| sp.last.in?(report_params[:sub_programs]) }
```
ttoomey: Just checking: we're confident that the sub_program_list ids are of the same type as report_params[:sub_programs] (strings / ints)?

This is totally fine as is; might be a little clearer to use filter instead of select and name the filtered vars. 
```suggestion
          @included_sub_programs = sub_program_list.filter { |_name, sp_id| sp_id.in?(report_params[:sub_programs]) }
```

app/controllers/reports/match_progress_controller.rb
```diff
@@ -85,20 +85,26 @@ def step_names(sub_program_id)
...
     end
-    helper_method :sub_programs
 
     def sub_program_list
-      @sub_program_list ||= sub_programs.map do |project_name, sub_project_name, id|
+      @sub_program_list ||= sub_programs.map do |sp|
         [
-          [project_name, sub_project_name].join('|'),
-          id,
+          [sp[:program], sp[:sub_program], sp[:route]].compact_blank.join('|'),
```
ttoomey, line 104: It looks like using compact here introduces ambiguity around which names are for which type of entity. Maybe that doesn't matter unless something is parsing them out (for example this would be incorrect `program, sub_program, route = name.split('|')`)

ttoomey (approved):
looks good!

app/controllers/reports/match_progress_controller.rb
eanders: Thanks, I definitely like that syntax better.

And yes, double checked, ints all the way down.

app/controllers/reports/match_progress_controller.rb
eanders: This is just used for display, I'd rather turn this into a sentence or more descriptive text, but didn't want to mess with the wording too much until we confirm this works.

PR #926, opened 2025-03-25
clean up fuzz for TC HAT dob so it alwasy falls within the age range

dtgreiner:
## _Merging this PR_

## Description
The TC_HAT assessment fuzzed dates of birth were able to fall outside of the expected date range. This update lets the fuzzed dob calculation to corrrectly account for the date fuzzing on the first and last years in the range.

## Type of change
Bug fix

### DIFF
--- a/spec/models/deidentified_tc_hat_spec.rb
+++ b/spec/models/deidentified_tc_hat_spec.rb
@@ -1,0 +1,37 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DeidentifiedTcHat, type: :model do
+  describe 'Fuzzed dates of birth' do
+    let!(:assessment) { DeidentifiedTcHat.new }
+    let!(:age_ranges) { assessment.send(:age_ranges) }
+    it 'fall within the expected date range' do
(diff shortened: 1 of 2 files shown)


spec/models/deidentified_tc_hat_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        # calculate the earliest and latest date of birth based on the age range
+        earliest_dob = (Date.current - v[:range].last.years)
+        latest_dob = (Date.current - v[:range].first.years)
+
+        # Date of birth is being fuzzed. We are running this test 10,000 times
+        # to get a probablily that we hit as many instances of the range as possible
+        # and to account for the randomness of the fuzzing.
+        # This is not a perfect test, but it should give us a good indication
+        # that the fuzzing is working as expected.
+        10_000.times do
```
eanders, line 31: This is probably fine, since this is probably fast, but could we figure out how to just check the edges?  Then we should only need a handful of checks. I'm guessing that's probably not possible.

eanders (approved):
Thanks @dtgreiner !

spec/models/deidentified_tc_hat_spec.rb
dtgreiner: The tests were running quickly locally, and adjusting the dob was hitting a failure each time. I could extract the fuzzed_dob calculations out of the `date_of_birth` method to allow specifying the age we are looking for  and then test on that. It would add a little extra unnecessary code in terms of what is needed for the applicaiton but would allow for better testing.

PR #927, opened 2025-04-02
CAS analytics data

eanders:
## _Merging this PR_

## Description
* Scripted export of CAS CE referral data for OP Analytics.
* Depends on https://github.com/greenriver/hmis-warehouse/pull/5263 being merged/deployed first

## Type of change
* New feature


app/models/warehouse/analytics/client.rb
```diff
@@ -0,0 +1,38 @@
...
+          hmis_client_id = client.project_client.id_in_data_source
+          next unless hmis_client_id.present?
+
+          batch << new(
+            id: client.id,
+            client_id: hmis_client_id,
+            calculated_first_homeless_night: client.calculated_first_homeless_night,
+            calculated_last_homeless_night: client.calculated_last_homeless_night,
+            created_at: client.created_at,
+            updated_at: client.updated_at,
```
dtgreiner, line 31: Is this all the client data we need here? I'm assuming yes and that the client_id is going to tie to the additional client data (e.g. name, dob, etc...) but thought it worth checking on.

app/models/warehouse/analytics/opportunity.rb
```diff
@@ -0,0 +1,34 @@
...
+      ::Opportunity.joins(:voucher).
+        preload(voucher: :unit).
+        distinct.
+        find_in_batches(batch_size: 1_000) do |opportunities|
+        batch = []
+        opportunities.each do |opportunity|
+          batch << new(
+            id: opportunity.id,
+            opportunity_category_id: opportunity.voucher.sub_program_id,
+            unit_id: opportunity.voucher.unit_id,
```
dtgreiner, line 24: Do we need to bring in the `unit_id` if we aren't bringing in the Unit table? It does add a uniqueness for units with the same name

dtgreiner (approved):
Looks good to me! I have a couple questions, but nothing that would be a blocker.

app/models/warehouse/analytics/client.rb
eanders: My thinking is that we'll probably need other eligibility criteria, but I'm not clear what yet.  For demographics, we should be able to link up to the existing warehouse data and for PII the existing warehouse client PII table.

app/models/warehouse/analytics/opportunity.rb
eanders: Exactly, I'm brining `unit_id` because there is no uniqueness constraint on the name.  Eventually, we may want additional unit information, but until we see what analysis people want to do, I figured less is more.

app/models/warehouse/analytics.rb
```diff
@@ -0,0 +1,11 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
```
ttoomey, line 8: Maybe leave a little documentation to help explain this


```suggestion
# The Analytics module supports export of CAS to our analytics store. These tables are consumed view views in the warehouse analytics schema
#
# Data Flow:
# CAS::Analytics --> CAS db connection --> analytics views (warehouse)  --> DBT (external) 
```

app/models/warehouse/analytics/cas_user.rb
```diff
@@ -0,0 +1,30 @@
...
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class Warehouse::Analytics::CasUser < ::Warehouse::Base
+  self.table_name = 'cas_analytics_cas_users'
+  def self.sync!
+    transaction do
+      delete_all
```
ttoomey: For all of these tables, if they are significant in size, TRUNCATE is probably more appropriate than DELETE (should be less costly in terms of IO)

ttoomey (approved):
Looks good!

PR #947, opened 2025-05-22
7721 improve permission checks on non-hmis client and assessment

ttoomey:
## _Merging this PR_

## Description
fixes to permission checks on clients and assessments
* fix permissions for identified/de-identified clients in assessment controller client and assessment loaders
* fix permissions in client.remote_visible_to
* correct permissions in the visible_to scope in the client child classes (DeidentifiedClient, IdentifiedClient, and Imported Client). Also DRY up the agency permission check
* remove unused visible_by and editable_by from NonHmisAssessment
* upgrade rubocop, it was crashing while trying to lint the specs

## Type of change
Bug fix, tests

- [x] I have provided testing instructions in this PR or the related issue (or not applicable)

### DIFF
--- a/app/models/deidentified_client.rb
+++ b/app/models/deidentified_client.rb
@@ -50,6 +50,7 @@ class DeidentifiedClient < NonHmisClient
   end
 
   def editable_by?(user)
+    return true if user.can_manage_all_deidentified_clients?
     return true if user.can_manage_deidentified_clients?
     return true if pathways_enabled? && user.can_enter_deidentified_clients?
     return user.agency_id == agency_id if user.can_enter_deidentified_clients?
--- a/app/models/deidentified_client.rb
+++ b/app/models/deidentified_client.rb
@@ -14,19 +14,8 @@ class DeidentifiedClient < NonHmisClient
   scope :visible_to, ->(user) do
     if user.can_edit_all_clients? || user.can_manage_all_deidentified_clients?
       all
-    elsif user.can_view_all_covid_pathways?
-      covid_ids = where(id: NonHmisAssessment.limitable_pathways.select(:non_hmis_client_id)).pluck(:id)
-      agency_associated_ids = where(
-        arel_table[:agency_id].eq(nil).
-        or(arel_table[:agency_id].eq(user.agency.id)),
-      ).pluck(:id)
-
-      where(id: covid_ids + agency_associated_ids)
     else
-      where(
-        arel_table[:agency_id].eq(nil).
-        or(arel_table[:agency_id].eq(user.agency.id)),
-      )
+      visible_through_user_agency(user)
     end
   end
 
--- a/app/models/imported_client.rb
+++ b/app/models/imported_client.rb
@@ -59,6 +59,14 @@ class ImportedClient < NonHmisClient
     where(where)
   end
 
+  scope :visible_to, ->(user) do
+    if user.can_edit_all_clients? || user.can_manage_imported_clients?
+      all
+    elsif visible_through_user_agency(user)
+      visible_through_user_agency(user)
+    end
+  end
+
   def editable_by?(user)
     return true if user.can_manage_imported_clients?
 
--- a/app/models/imported_client.rb
+++ b/app/models/imported_client.rb
@@ -62,7 +62,7 @@ class ImportedClient < NonHmisClient
   scope :visible_to, ->(user) do
     if user.can_edit_all_clients? || user.can_manage_imported_clients?
       all
-    elsif visible_through_user_agency(user)
+    else
       visible_through_user_agency(user)
     end
   end
--- a/app/models/non_hmis_assessment.rb
+++ b/app/models/non_hmis_assessment.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 ###
 # Copyright 2016 - 2025 Green River Data Analysis, LLC
 #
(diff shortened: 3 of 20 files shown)


app/models/non_hmis_client.rb
```diff
@@ -31,10 +31,8 @@ class NonHmisClient < ApplicationRecord
     where(available: true)
   end
 
-  scope :visible_to, ->(user) do
```
ttoomey, line 34: moved visible_to to child classes

app/models/non_hmis_assessment.rb
```diff
@@ -29,14 +31,6 @@ class NonHmisAssessment < ActiveRecord::Base
     where(type: limited_assessment_types)
   end
 
-  scope :visible_by, ->(user) do
-    where(non_hmis_client_id: NonHmisClient.visible_to(user).select(:id))
-  end
-
-  scope :editable_by, ->(user) do
-    where(agency_id: user.agency_id)
-  end
```
ttoomey, line 38: these appear unused

app/models/imported_client.rb
```diff
@@ -59,6 +59,14 @@ class ImportedClient < NonHmisClient
     where(where)
   end
 
+  scope :visible_to, ->(user) do
+    if user.can_edit_all_clients? || user.can_manage_imported_clients?
```
ttoomey, line 63: can_manage_imported_clients was missing previously as it was using NonHmisClient#visible_to

app/models/non_hmis_assessment.rb
```diff
@@ -140,11 +134,11 @@ def set_non_hmis_client_availability
   end
 
   def pathways_v3?
-    type.include?('PathwaysVersionThree')
```
ttoomey, line 143: safe navigation needed for test factory to work

spec/models/client_spec.rb
```diff
@@ -618,4 +618,199 @@
...
+    let!(:basic_user_in_other_agency) { create(:user, agency: other_agency, roles: [basic_role]) }
+
+    let!(:deid_manager_role) { create(:role, name: 'remote_client_test_deid_manager', can_manage_all_deidentified_clients: true) }
+    let!(:deid_viewer) { create(:user, agency: user_agency, roles: [deid_manager_role]) }
+
+    # ASSUMPTION: Role model has can_manage_all_identified_clients attribute
+    let!(:id_manager_role) { create(:role, name: 'remote_client_test_id_manager', can_manage_all_identified_clients: true) }
+    let!(:identified_viewer) { create(:user, agency: user_agency, roles: [id_manager_role]) }
+
+    # ASSUMPTION: Role model has can_manage_imported_clients attribute
```
eanders: Are the se comments just slop?  Seems like we should know what attributes `Role` has.

ttoomey: oh yeah, I think some of them can be pruned, I'll do that

spec/models/deidentified_client_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    let!(:user_with_covid_pathways_role) do
+      user = create(:user, agency: user_agency)
+      role = create(:role, name: 'covid_pathway_viewer',
+                           can_edit_all_clients: false,
+                           can_manage_all_deidentified_clients: false,
+                           can_view_all_covid_pathways: true)
+      user.roles << role
+      user
+    end
```
eanders: It looks like we're testing covid pathways (that no one uses any more) but not the standard `can_edit_diedentified_clients` which is what we just had issues with. Can we add that set of permissions to some tests?

eanders (changes requested):
I think the changes look good, but if we're going to add all these tests we should probably test the permissions that were triggered the original issue.
```
user.can_enter_identified_clients? 
user.can_enter_deidentified_clients?
```

ttoomey:
> However I'm not sure about the specific permission you mentioned in your feedback:
> user.can_enter_identified_clients? 
> user.can_enter_deidentified_clients?
> can_edit_diedentified_clients
> 
> The code change was to allow users with can_manage_all_deidentified_clients and can_manage_all_identified_clients create to client assessments. But maybe we need to include other permissions in the `visible_to` scope, such as 'can_enter_*' (I guess 'enter' means create records?)

@ttoomey that's pretty much it.  The `can_manage_all_*` methods are "administrative".  There are equivalent `can_enter_*` permissions that only allow you to edit the clients at your agency.

From the controller:
```
private def require_can_edit_this_client!
    not_authorized! unless @non_hmis_client&.editable_by?(current_user)
  end
```

From De-identified clients:
```
def editable_by?(user)
    return true if user.can_manage_deidentified_clients?
    return true if pathways_enabled? && user.can_enter_deidentified_clients?
    return user.agency_id == agency_id if user.can_enter_deidentified_clients?

    false
  end
```

Somewhat oddly the Identified client version looks somewhat different.  Then end result looks like it would be similar:
```
scope :editable_by, ->(user) do
    if user.can_edit_all_clients? || user.can_manage_all_identified_clients?
      all
    else
      return none unless user.can_manage_identified_clients? || user.can_enter_identified_clients?

      if pathways_enabled?
        all
      else
        where(agency_id: user.agency.id)
      end
    end
  end
```

We should add a test or two to confirm if you have `can_enter_*` that you can only edit clients at your agency.

eanders (approved):

ttoomey:
@eanders - looks like you edited my initial comment above to give feedback but I think I missed that, sorry about that

> We should add a test or two to confirm if you have can_enter_* that you can only edit clients at your agency.

Good idea, I'll do that, thanks

app/models/deidentified_client.rb
```diff
@@ -61,6 +50,7 @@ def self.assessment_type
   end
 
   def editable_by?(user)
+    return true if user.can_manage_all_deidentified_clients?
```
ttoomey, line 53: seems like can_manage_all_* should grant editable permission?

eanders: Makes sense.

eanders (approved):

PR #952, opened 2025-06-04
Alpine -> Debian

eanders:
## _Merging this PR_

## Description
* Converts docker images from alpine to debian
* Backports some changes for CI from the warehouse

## Type of change
* Dependency update

- [ ] I have performed a self-review of my code


.github/workflows/container_setup/action.yml
```diff
@@ -0,0 +1,133 @@
...
+          libicu-dev \
+          libmagic-dev \
+          libpq-dev \
+          libxml2-dev \
+          libxslt1-dev \
+          libyaml-dev \
+          nodejs \
+          poppler-utils \
+          postgresql-client-16 \
+          postgresql-client-17 \
```
yakloinsteak, line 84: Why do we need two versions of postgres? Is there a possibility of not installing python? Just trying to help speed up builds. Maybe freetds-dev and the fonts can go too?

.github/workflows/rails_tests.yml
```diff
@@ -41,42 +41,29 @@ jobs:
...
-          gem install bundler --version=2.6.3
-          bundle config set --local without 'production staging development'
-          bundle install --jobs 4 --retry 3
-
       - name: 'App setup'
         run: |
           cp config/secrets.yml.sample config/secrets.yml
           mkdir app/assets/stylesheets/theme/styles
           touch app/assets/stylesheets/theme/styles/_variables.scss
           cp .rspec.travis .rspec
```
yakloinsteak: travis? Could use a renaming probably.

docker/app/Dockerfile
```diff
@@ -47,37 +57,88 @@ ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US.UTF-8
 # Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
-ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-browser
+ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium
 ENV GROVER_NO_SANDBOX=true
-ENV CHROMIUM_PATH=/usr/bin/chromium-browser
+ENV CHROMIUM_PATH=/usr/bin/chromium
```
yakloinsteak: Just wanted to double check that we're using puppeteer and chromium over here. Those dependencies drastically increase the image sizes and build times.

docker/app/Dockerfile
```diff
@@ -47,37 +57,88 @@ ENV LANG=en_US.UTF-8
...
 
 # Allow tls connection to database with verification
 COPY docker/app/us-east-1-bundle.pem  /etc/ssl/certs/us-east-1-bundle.pem
 # Deprecated and soon to expire
 COPY docker/app/rds-combined-ca-bundle.pem /etc/ssl/certs/rds-combined-ca-bundle.pem
```
yakloinsteak: We haven't needed this one for a while I think.

docker/app/Dockerfile
```diff
@@ -47,37 +57,88 @@ ENV LANG=en_US.UTF-8
...
+  fonts-liberation \
+  freetds-dev \
+  git \
+  gnupg-agent \
+  gpg \
+  imagemagick \
+  libcurl4 \
+  libcurl4-openssl-dev \
+  libffi-dev \
+  libfreetype-dev \
```
yakloinsteak, line 96: Our pocketbook, build times, and boot times will thank us if we can reduce this set of packages. Maybe Dave could take a pass?

docker/docker-compose.yml
```diff
@@ -1,11 +1,11 @@
...
-  image: cas:${TARGET:-dev-build}-1.1
+  image: cas:${TARGET:-dev-build}-2.0
   build:
     context: .
     target: ${TARGET:-dev-build}
     dockerfile: docker/app/Dockerfile
     args:
-      BUILD_TAG: 3.3.7-alpine3.20
+      BUILD_TAG: 3.3.8-slim-bookworm
       PG_MAJOR: '12'
```
yakloinsteak: Looked quickly, and these PG args I think aren't used. I'd remove any that aren't in use.

yakloinsteak (approved):
Looks fine, though I highly recommend we make some attempt to reduce image sizes if possible.

docker/app/Dockerfile
eanders: Thanks, this was just a copy/paste from the warehouse.  I'm not seeing any reason we need chrome or related bits.  Removed.

docker/app/Dockerfile
eanders: removed

docker/app/Dockerfile
eanders: removed, pretty sure we don't need these, I'm not seeing any PDF generation in ere

docker/docker-compose.yml
eanders: PG_MAJOR removed, were there others that should go too?

yakloinsteak: They're only needed if there is an ARG line in the Dockerfile to match.

docker/app/Dockerfile
```diff
@@ -45,39 +55,80 @@ ENV RAILS_LOG_TO_STDOUT=true
...
+  libmagic-dev \
+  libmagic1 \
+  libpq-dev \
+  libx11-6 \
+  libxext6 \
+  libxml2-dev \
+  libxrender1 \
+  libxslt1-dev \
+  libyaml-dev \
+  locales-all \
```
ttoomey, line 112: If we're trying to be more frugal, looks like , looks like locales-all is north of 200MB. We could do:
```suggestion
  locales \
```
but then we'd probably also need to set the locale too, possibly with this (not tested)
```
RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen \
  && locale-gen \
  && update-locale LANG=en_US.UTF-8
```

Chromium is also big but looks like it's been removed. Also awscli is 95MB for some reason but we probably need that for something. Looking at

```bash
dpkg-query -Wf='${Installed-Size}\t${Package}\n' | sort -n
```

ttoomey (approved):
Looks great, thanks!

PR #960, opened 2025-06-16
Entrypoint syntax fixes

yakloinsteak:
- Fixes shell syntax errors
- Lints the script

## Description
I think the move from alpine left us with different shell behavior, and we usually use double brackets for comparisons with the double equals. The single brackets are for operators like `-eq`. Figured bash is more powerful anyway, so changes to that.

## Type of change
bugfix

## Testing
I ran this in a container to test before and after the change:

```
EKS=true ./docker/app/entrypoint.sh echo foo
```

### DIFF
--- a/docker/app/entrypoint.sh
+++ b/docker/app/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/bash
 
 # Exit on any error
 # set -e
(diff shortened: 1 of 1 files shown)


dtgreiner (approved):
Looks good to me! Does this need updated in the warehouse too?

yakloinsteak:
The warehouse didn't have this exact same syntax problem, but porting that one to bash with the double brackets and two equal symbols for equality would be good. Do you have time to take that on?

PR #976, opened 2025-07-31
Admin analytics access

eanders:
## _Merging this PR_

## Description
* Fakes a contact record for every match for everyone who can reject or approve matches so they can have access in OP Analytics for reporting

## Type of change
* New feature

- [ ] I have performed a self-review of my code

### DIFF
--- a/app/models/warehouse/analytics/referral_contact.rb
+++ b/app/models/warehouse/analytics/referral_contact.rb
@@ -19,7 +19,6 @@ class Warehouse::Analytics::ReferralContact < ::Warehouse::Base
         batch = []
         match_contacts.each do |match_contact|
           batch << new(
-            # id: match_contact.id,
             email: match_contact.contact.email,
             referral_id: match_contact.match.id,
             contact_id: match_contact.contact_id,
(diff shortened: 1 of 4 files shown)


app/models/warehouse/analytics/referral_contact.rb
```diff
@@ -19,7 +19,6 @@ def self.sync!
         batch = []
         match_contacts.each do |match_contact|
           batch << new(
-            id: match_contact.id,
```
eanders, line 22: Not sure why we had this specified, but we can't rely on it now that we're faking contacts.

dtgreiner:
Am I making the correct assumptions here as to why we are making this change?
- match admins currently cannot report on matches that they are not an official contact for
- we're not exporting any other match admin data to superset which would allow that viewability
- we are including the match admins as a spoofed contact on the matches they admin (only in the superset data) to allow reporting on the matches

eanders:
> Am I making the correct assumptions here as to why we are making this change?
> 
> * match admins currently cannot report on matches that they are not an official contact for
> * we're not exporting any other match admin data to superset which would allow that viewability
> * we are including the match admins as a spoofed contact on the matches they admin (only in the superset data) to allow reporting on the matches

That's why!  
We chose to use the match-contact association as the filter for which matches you can report on in OP Analytics, but didn't account for people who don't participate in matches, but need a means of reporting on them.  The access is row-level, so adding them here seemed the most logical since they can also see these matches when looking in CAS.

dtgreiner (approved):

PR #987, opened 2025-08-22
Persisted Search Queries

eanders:
## _Merging this PR_

## Description
* Port of protected PII searching from the warehouse
* Adjusts the search pages to use post instead of get to prevent leaking PII in the URL
* Persists search queries to the database
* Re-uses persisted search queries for identical searches
* Applies to:
  - Clients (available/unavailable)
  - Deidentified clients
  - Identified clients
  - Imported clients
  - Active matches
  - Closed matches
  - Admin users

## Type of change
* New feature

- [x] I have provided testing instructions in this PR or the related issue (or not applicable)


Gemfile
```diff
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 gem 'activerecord-import'
 gem 'rack'
 gem 'pg', '~> 1.1'
-gem 'rails', '~> 7.1.0'
+gem 'rails', '~> 7.1.5.2'
```
ttoomey: seems like there's a merge issue

app/controllers/admin/users_controller.rb
```diff
@@ -17,13 +19,30 @@ class UsersController < ApplicationController
...
-        @inactive_users = User.inactive
+      # Handle search queries
+      handle_search_query
+
+      filter_data
+    end
+
+    def search
+      @search_query = ClientSearchQuery.find(params[:id])
+      return handle_invalid_query('Search query not found') if @search_query.nil?
```
ttoomey: this case is unreachable because `find` will raise. Probably use `find_by(id: params[:id])` to avoid raise

app/controllers/admin/users_controller.rb
```diff
@@ -169,5 +194,21 @@ def sort_direction
...
+
+      search_params = ActionController::Parameters.new({ q: params[:q] })
+      permitted_params = ClientSearchQuery.permit_params(search_params)
+      return unless permitted_params.present?
+
+      @search_query = ClientSearchQuery.find_or_create_by_params(permitted_params, user: current_user)
+      return if @search_query.errors.any?
+
+      # Redirect to the search query URL if this is a GET request
+      redirect_to admin_user_search_query_path(@search_query) if request.get?
```
ttoomey: maybe this fixes the brakeman warning
```suggestion
      redirect_to admin_user_search_query_path(@search_query) if request.get? || request.head?
```

app/models/client_search_query.rb
```diff
@@ -0,0 +1,93 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class ClientSearchQuery < ApplicationRecord
```
ttoomey, line 26: call out in a comment that this model contains persisted PII?

app/controllers/admin/users_controller.rb
```diff
@@ -17,13 +19,30 @@ class UsersController < ApplicationController
...
     def index
-      # search
-      if params[:q].present?
-        @users = user_scope.text_search(params[:q])
-        @inactive_users = User.inactive.text_search(params[:q])
-      else
-        @users = user_scope
-        @inactive_users = User.inactive
+      # Handle search queries
+      handle_search_query
```
ttoomey, line 23: avoid double render?
```suggestion
      handle_search_query
      return if performed?
```

app/controllers/clients_controller.rb
```diff
@@ -19,14 +19,33 @@ class ClientsController < ApplicationController
 
   # GET /hmis/clients
   def index
+    # Handle search queries
+    handle_search_query
```
ttoomey, line 23:
```suggestion
    handle_search_query
    return if performed?

```

app/controllers/non_hmis_clients_controller.rb
```diff
@@ -13,18 +15,72 @@ class NonHmisClientsController < ApplicationController
...
   before_action :load_contacts, only: [:new, :edit]
-  before_action :set_active_filter, only: [:index]
+  before_action :set_active_filter, only: [:index, :search]
   before_action :find_match, only: [:current_assessment_limited]
 
+  helper_method :non_hmis_client_search_queries_path
+
   def index
+    # Handle search queries
+    handle_search_query
```
ttoomey, line 25:
```suggestion
    handle_search_query
    return if performed?
    
```

app/controllers/match_list_base_controller.rb
```diff
@@ -17,21 +19,57 @@ def initialize
...
 
   def index
-    # search
-    @matches = if params[:q].present?
-      match_scope.text_search(params[:q])
-    else
-      match_scope
-    end
+    # Handle search queries
+    handle_search_query
```
ttoomey, line 23:
```suggestion
    handle_search_query
    return if performed?
    
```

app/views/clients/_filter.haml
```diff
@@ -10,8 +10,7 @@
     .o-filtration__header
       %h3.client__filter-header Filter clients
 
-    = form_tag(nil, method: "get", class: :jFilter) do
-      -# preserve sort and search for filtering
+    = form_tag(request.original_url, method: "get", class: :jFilter) do
```
ttoomey: I think original_url includes query params, so that mixes params from get and post.
```suggestion
    = form_tag(url_for(only_path: true), method: "get", class: :jFilter) do
```

app/controllers/admin/users_controller.rb
```diff
@@ -169,5 +194,21 @@ def sort_direction
...
+      return unless permitted_params.present?
+
+      @search_query = ClientSearchQuery.find_or_create_by_params(permitted_params, user: current_user)
+      return if @search_query.errors.any?
+
+      # Redirect to the search query URL if this is a GET request
+      redirect_to admin_user_search_query_path(@search_query) if request.get?
+    rescue ActiveRecord::RecordInvalid
+      # Handle validation errors gracefully
+    end
```
ttoomey: we don't use the bang variants so I'm unclear on what the rescue does here (and in other controllers)

app/controllers/non_hmis_clients_controller.rb
```diff
@@ -127,12 +195,12 @@ def sorter
     if @column.blank?
       if pathways_enabled?
         if can_manage_identified_clients?
-          @column = 'non_hmis_clients.assessment_score'
+          @column = 'assessment_score'
```
ttoomey, line 200: unclear why tablename is removed

ttoomey (commented):
Looks good, there's a lot of changes the test coverage adds confidence. Thanks for that.

Suggestions:
* We don't have rack-attack in the project. In the warehouse we use rack-attack to protect against brute forcing search queries (and other sensitive end points). I think it's important to include it here
*  Seems like we aren't pruning old search queries (see warehouse CleanupClientSearchQueriesTask). That reduces the risk of exposure
* There's churn in the structure file that seems unrelated to the migration. And looks like some unrelated changes have crept in (rails upgrade or something)

eanders:
> Looks good, there's a lot of changes the test coverage adds confidence. Thanks for that.
> 
> Suggestions:
> 
> * We don't have rack-attack in the project. In the warehouse we use rack-attack to protect against brute forcing search queries (and other sensitive end points). I think it's important to include it here
> * Seems like we aren't pruning old search queries (see warehouse CleanupClientSearchQueriesTask). That reduces the risk of exposure
> * There's churn in the structure file that seems unrelated to the migration. And looks like some unrelated changes have crept in (rails upgrade or something)

Thanks @ttoomey !

* I'll add rack attack in a new PR, probably against rails 7.2 as this is pretty large as is.
* Good call on the pruning, I'll add that and address your other in-line comments in this PR.
* I just tried rebuilding the structure file from a fresh test DB based off release-88.  Not sure why, but it comes out the same.  This should only be of consequence for new installations, which we should not have at this time, so leaving it for now.

app/controllers/admin/users_controller.rb
eanders: Pretty sure that was slop.

app/controllers/non_hmis_clients_controller.rb
eanders: That was breaking the search/sort.  Removing it seemed to make it happy.

ttoomey (approved):
Thanks Elliot, looks good!
Fine to do in another PR but I think we shouldn't actually deploy to prod without rack-attack in place

PR #1000, opened 2025-09-10
Work around for sorting

eanders:
## _Merging this PR_

## Description
* Fixes a bug where we needed to exclude some columns to be able to sort, but needed them to "build" missing assessments.

## Type of change
* Bug fix

### DIFF
--- a/app/controllers/non_hmis_clients_controller.rb
+++ b/app/controllers/non_hmis_clients_controller.rb
@@ -17,6 +17,7 @@ class NonHmisClientsController < ApplicationController
   before_action :load_contacts, only: [:new, :edit]
   before_action :set_active_filter, only: [:index, :search]
   before_action :find_match, only: [:current_assessment_limited]
+  around_action :with_skip_build_assessment_if_missing, only: [:index]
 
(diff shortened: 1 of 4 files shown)


app/controllers/non_hmis_clients_controller.rb
```diff
@@ -28,6 +29,15 @@ def index
...
   end
 
+  # This allows us to skip the build_assessment_if_missing method in the non_hmis_client model
+  # which calls some attributes that are explicitly not loaded in the index action so sort will work
+  private def with_skip_build_assessment_if_missing
+    Current.skip_build_assessment_if_missing = true
+    yield
+  ensure
+    Current.skip_build_assessment_if_missing = nil
+  end
```
dtgreiner, line 39: Will this always be starting as nil or would it be worth storing the current value as a variable and resetting it to that afterwards?

eanders: Always `nil`.  We only set this in the tests at this point.
Actually, we only set this here, the tests use the class attribute.

dtgreiner (approved):

PR #1040, opened 2025-12-23
Affordance for location specific notes on rules

eanders:
## _Merging this PR_

## Description
* Adds a note for the Bedroom rule that shows up on the add voucher rules page
* Implements a structure such that extra notes for rules can be added and displayed in appropriate locations
* Adjusts translation interface to use a text area to support multi-line translations (matches warehouse)

## Type of change
* New feature

### DIFF
--- a/app/models/rules/bedroom.rb
+++ b/app/models/rules/bedroom.rb
@@ -11,6 +11,20 @@ class Rules::Bedroom < Rule
     'Matches clients who require at least the number of bedrooms specified in "Minimum number of bedrooms required" as indicated on an assessment.'
   end
 
+  # supports markdown
+  def selection_note(context: nil)
+    return unless context == :voucher
+
+    # Note for the voucher requirements page, which appears to be have in the reverse of the expected logic.
+    note = <<~NOTE
+      ### Note
+      If this voucher is attached to a 2 bedroom unit, to match a client who requires 2 or more bedrooms, choose "Can't have minimum number of bedrooms" with "One" selected as the number of bedrooms.
+
+      For a 3 bedroom unit, choose"Can't have minimum number of bedrooms" with "Two".
+    NOTE
+    Translation.translate(note)
+  end
+
   def variable_requirement?
     true
   end
--- a/app/models/rules/bedroom.rb
+++ b/app/models/rules/bedroom.rb
@@ -20,7 +20,7 @@ class Rules::Bedroom < Rule
       ### Note
       If this voucher is attached to a 2 bedroom unit, to match a client who requires 2 or more bedrooms, choose "Can't have minimum number of bedrooms" with "One" selected as the number of bedrooms.
 
-      For a 3 bedroom unit, choose"Can't have minimum number of bedrooms" with "Two".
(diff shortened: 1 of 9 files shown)


app/models/rules/bedroom.rb
```diff
@@ -11,6 +11,20 @@ def description
     'Matches clients who require at least the number of bedrooms specified in "Minimum number of bedrooms required" as indicated on an assessment.'
```
ttoomey, line 11: I'm not super familiar with how these rules work and am probably not understanding but is this description supposed to summarize the behavior of `clients_that_fit()`

Seems like `requirement.positive` would need to be accounted for in the description. Also when we say "at least" I'm not sure that matched the 'gt()` condition in the filter.

Can you confirm the description and filter logic are accurate?

```ruby
  def description(requirement = nil)
    if requirement.positive
      'Matches clients who require less than the number of bedrooms specified in "Minimum number of bedrooms required" as indicated on an assessment.'
    else
      # maybe the filter should use gteq() if it is "at least"
      'Matches clients who require at least the number of bedrooms specified in "Minimum number of bedrooms required" as indicated on an assessment.'
    end
  end
 ```

eanders: Good question.  At the moment the `selection_note` appears when you choose to use the rule, but shows the same content for if you choose "must" (positive) or "can't" (negative).  I suppose we could extend the note text to describe the "can't" scenario, but as it is right now, only the specific scenarios noted are causing trouble when setting up the rule correctly.

Looks like I have a typo though, so I'll push a fix for the missing space.

eanders: Oh, I see, you're suggesting we change the main description.  Those are shown on a page that lists all rules and doesn't go into details about the positive or negative choices.

ttoomey: yeah, ignore my feedback if it seems like the description isn't used in a context that matters. Just flagging that, as written, the description doesn't seem to accurately describe the behavior of the filter in all cases

ttoomey: To try and explain the discrepancy a little more: 

If I set this to 2 bedrooms, it will match all clients with a requirement of 2 bedrooms or less. So a client that requires 1 bedroom will match, assuming request.positive is true.

However the language is "Matches clients who require at least the number of bedrooms specified". So I would expect the filter to match all clients who require 2 or more bedrooms (a client requiring 3 bedrooms should match.

ttoomey: Also this issue is outside of the scope of this PR and probably shouldn't block it; I'll approve

ttoomey (approved):

app/models/rules/bedroom.rb
eanders: Thanks.  The reason this extra description was needed is that this particular rule can be applied in multiple locations and has somewhat unexpected behavior in some of them.  The note is supposed to make it clear how to use it in the one situation it seems off, but it's also possible something else is wrong.  Given that it's been working elsewhere I'm not tempted to adjust that at the moment, but good feedback, thanks!

PR #1060, opened 2026-03-04
Match History Step Names

dtgreiner:
## _Merging this PR_

## Description

Use the event timestamp to determine the active decision when the event took place (when the decision is not tied to the event). Use this to populate the step_name in the match history.

## Type of change

New feature

### DIFF
--- a/app/models/client_opportunity_match.rb
+++ b/app/models/client_opportunity_match.rb
@@ -508,6 +508,15 @@ class ClientOpportunityMatch < ApplicationRecord
     @current_decision = nil
   end
 
+  def decision_active_at(timestamp)
+    return nil unless timestamp.present?
+
+    initialized_decisions
+      .where('match_decisions.updated_at <= ?', timestamp)
+      .order(updated_at: :desc)
+      .first || initial_decision
+  end
+
   def add_default_contacts!
     add_default_dnd_staff_contacts!
     add_default_housing_subsidy_admin_contacts!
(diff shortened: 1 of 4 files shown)


dtgreiner:
@eanders here's a quick pass. Happy to rework it if we want to dig deeper into the decisions for these event types.

app/models/client_opportunity_match.rb
```diff
@@ -508,6 +508,15 @@ def clear_current_decision_cache!
     @current_decision = nil
   end
 
+  def decision_active_at(timestamp)
+    return nil unless timestamp.present?
+
+    initialized_decisions
+      .where('match_decisions.updated_at <= ?', timestamp)
```
eanders, line 515:
```suggestion
      .where(updated_at: ..timestamp)
```

app/models/client_opportunity_match.rb
```diff
@@ -508,6 +508,15 @@ def clear_current_decision_cache!
     @current_decision = nil
   end
 
+  def decision_active_at(timestamp)
```
eanders, line 511: This is going to create a bunch of N+1s.  Is there no way to infer the step from the notifications for a give step?  That's probably pretty hard....

Not sure I like relying on time-based step determination since you can backup a match and some steps are probably decline steps.

How hard would it be to pass the step to the notification and store the step_id so we have a direct linkage going forward?

PR #1061, opened 2026-03-06
Match History - Include more decision step data where possible

dtgreiner:
## _Merging this PR_

## Description

Propagates decision_id from decisions to notification delivery events so events are tied to the decision that created them. 

The backfill only sets decision_id when a single decision in the match sends that notification type, avoiding overlap and incorrect assignment. 

OnBehalfOf logic is centralized using notify_on_behalf_of? and skip_notify_on_behalf_of_when_canceled? instead of per-decision overrides. Redundant create_for_match! implementations in notification subclasses are removed in favor of the base implementation.

## Type of change

New Feature, Code Cleanup

- [X] I have added spec tests (or not applicable)

### DIFF
--- a/app/models/notifications/eight/match_canceled.rb
+++ b/app/models/notifications/eight/match_canceled.rb
@@ -13,12 +13,8 @@ module Notifications::Eight
       [:contacts]
     end
 
-    def self.create_for_match! match
-      contact_types_for_notification.each do |contact_type|
-        match.send(contact_type).each do |contact|
-          create! match: match, recipient: contact
-        end
-      end
+    def self.create_for_match!(match, decision_id: nil)
+      Notifications::Base.instance_method(:create_for_match!).bind(self).call(match, decision_id: decision_id)
     end
   end
 end
--- a/app/models/notifications/five/match_canceled.rb
+++ b/app/models/notifications/five/match_canceled.rb
@@ -13,12 +13,8 @@ module Notifications::Five
       [:contacts]
     end
 
-    def self.create_for_match! match
-      contact_types_for_notification.each do |contact_type|
-        match.send(contact_type).each do |contact|
-          create! match: match, recipient: contact
-        end
-      end
+    def self.create_for_match!(match, decision_id: nil)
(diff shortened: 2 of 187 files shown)


app/models/notifications/eight/match_canceled.rb
```diff
@@ -13,12 +13,8 @@ def self.contact_types_for_notification
...
     end
 
-    def self.create_for_match! match
-      contact_types_for_notification.each do |contact_type|
-        match.send(contact_type).each do |contact|
-          create! match: match, recipient: contact
-        end
-      end
+    def self.create_for_match!(match, decision_id: nil)
+      Notifications::Base.instance_method(:create_for_match!).bind(self).call(match, decision_id: decision_id)
```
eanders, line 17: Why is this one different?  As far as I can tell the original is the same as those above.

app/models/notifications/five/match_canceled.rb
```diff
@@ -13,12 +13,8 @@ def self.contact_types_for_notification
       [:contacts]
     end
 
-    def self.create_for_match! match
-      contact_types_for_notification.each do |contact_type|
-        match.send(contact_type).each do |contact|
-          create! match: match, recipient: contact
-        end
-      end
+    def self.create_for_match!(match, decision_id: nil)
```
eanders, line 16: Same here.  Maybe because `contact_types_for_notification` returns `[:contacts]` which is somehow special?

app/models/notifications/eight/match_canceled.rb
dtgreiner: these MatchCanceled classes are inheriting from `Notifications::MatchCanceled` and not `Notifications:Base`. `Notifications:MatchCanceled` overrides this method and these were overriding the override to work like the base version.

dtgreiner: This was an attempt to keep the the change limited to methods affected by the addition of the decision_id and not letting the scope of the update get out of hand. We could change the MATCH::MatchCanceled methods here back to the original and add the decision_id, but this way consolidated the logic into one location.

eanders (approved):

PR #1079, opened 2026-04-14
fix: branch 9098 auth checks

ttoomey:
see GH issue

### DIFF
--- a/app/controllers/opportunities_controller.rb
+++ b/app/controllers/opportunities_controller.rb
@@ -261,8 +261,9 @@ class OpportunitiesController < ApplicationController
   private def query_string
     "%#{@query}%"
   end
+
   private def require_can_view_opportunities_or_can_add_vacancies!
-    can_view_opportunities? || can_add_vacancies?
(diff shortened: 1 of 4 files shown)


app/controllers/opportunities_controller.rb
```diff
@@ -261,8 +261,9 @@ def route_from_tab(tab_label)
   private def query_string
     "%#{@query}%"
   end
+
   private def require_can_view_opportunities_or_can_add_vacancies!
-    can_view_opportunities? || can_add_vacancies?
+    not_authorized! unless can_view_opportunities? || can_add_vacancies?
```
eanders, line 266: I wonder if this would read more clearly as:
```suggestion
    return true if can_view_opportunities? || can_add_vacancies?
    
    not_authorized!
```

eanders (approved):

app/controllers/opportunities_controller.rb
ttoomey: I agree that's more readable but probably not worth changing rn; the codebase seems to already use the short-form pattern for the `require_` methods

PR #1107, opened 2026-06-12
Vacancy Submission Cleanup

eanders:
## _Merging this PR_

## Description

* A first pass at aligning existing vacancy cleanup with the CAS styles
* Updates style guide
* Fixes for XSS issues
* Moves to stimulus controllers for Js

## Type of change

* Code clean-up

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,2 +1,0 @@
-web: bin/rails server
-js: yarn build --watch
--- a/app/controllers/style_guides_controller.rb
+++ b/app/controllers/style_guides_controller.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/boston-cas/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 class StyleGuidesController < ApplicationController
   include AjaxModalRails::Controller
 
@@ -15,4 +17,17 @@ class StyleGuidesController < ApplicationController
     @form = OpenStruct.new
   end
 
+  private def guide_routes
+    @guide_routes ||= {
+      summary: 'Summary',
+      dnd_match_review: 'DnD Match Review',
+      form: 'Form Elements',
+      icon_font: 'Icon Font',
+      pagination: 'Pagination',
+      stepped_progress: 'Stepped Progress',
+      tags: 'Tags',
+      typography: 'Typography',
+    }
+  end
+  helper_method :guide_routes
 end
--- a/app/models/vacancy_submission.rb
+++ b/app/models/vacancy_submission.rb
@@ -141,13 +141,13 @@ class VacancySubmission < ApplicationRecord
   private
 
   def required_draft_fields
-    errors.add(:base, 'Program is required') if program_id.blank?
-    errors.add(:base, 'Sub-program is required') if sub_program_id.blank?
+    errors.add(:program_id, :blank) if program_id.blank?
+    errors.add(:sub_program_id, :blank) if sub_program_id.blank?
     return if voucher?
 
-    errors.add(:base, 'Street address is required') if unit_address_street.blank?
-    errors.add(:base, 'City is required') if unit_address_city.blank?
-    errors.add(:base, 'State is required') if unit_address_state.blank?
-    errors.add(:base, 'Zip code is required') if unit_address_zip.blank?
+    errors.add(:unit_address_street, :blank) if unit_address_street.blank?
+    errors.add(:unit_address_city, :blank) if unit_address_city.blank?
+    errors.add(:unit_address_state, :blank) if unit_address_state.blank?
+    errors.add(:unit_address_zip, :blank) if unit_address_zip.blank?
   end
 end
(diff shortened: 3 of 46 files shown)


app/controllers/style_guides_controller.rb
```diff
@@ -15,4 +17,17 @@ def form
     @form = OpenStruct.new
   end
 
+  private def guide_routes
```
eanders, line 20: Style guide changes to align with the warehouse setup a bit better, and add badges to the summary page. Most of the views only changed to get a menu.

app/models/vacancy_submission.rb
```diff
@@ -6,12 +12,17 @@ class VacancySubmission < ApplicationRecord
   belongs_to :user, optional: true
   has_many :vacancy_submission_notes, dependent: :destroy
 
+  store_accessor :draft_data,
```
eanders, line 15: Added `store_accessor` to hopefully make it [easier to store and retrieve](https://api.rubyonrails.org/v7.2/classes/ActiveRecord/Store.html) data from the JSON blob

docker/app/entrypoint.dev.sh
```diff
@@ -26,5 +26,8 @@ bundle config --global set build.sassc --disable-march-tune-native
 bundle install --quiet || echo "bundle install failed"
 yarn install --silent --frozen-lockfile || echo "yarn install failed"
 
+yarn build || echo "yarn build failed"
```
eanders, line 29: Turns out we weren't auto building JS on change after the BS5 upgrade.

Procfile.dev
eanders, line 1: This was left over from when we used to start CAS with foreman.... pre docker

app/models/vacancy_submission.rb
```diff
@@ -45,28 +55,39 @@ def self.derive_is_voucher(sub_program)
...
-    return '—' if draft_data['is_voucher']
-    [
-      draft_data['unit_address_street'],
-      draft_data['unit_address_city'],
-      draft_data['unit_address_state'],
-      draft_data['unit_address_zip'],
-    ].compact.reject(&:blank?).join(', ').presence || '—'
+    return '—' if voucher?
+
+    [unit_address_street, unit_address_city, unit_address_state, unit_address_zip]
```
dtgreiner: It looks like the original was missing the unit number
```suggestion
    [unit_address_street, unit_address_unit_number, unit_address_city, unit_address_state, unit_address_zip]
```

app/views/vacancy_submissions/index.html.haml
```diff
@@ -15,7 +16,7 @@
 .table-responsive
   .c-card.c-card__flush.mb-4
     .c-card__content.p-0
-      %table.table.table-striped.mb-0
+      %table.table.table-hover.mb-0
```
dtgreiner, line 21: Are we using this elsewhere? The cursor that comes with the `table-hover` css makes me feel like the whole row should be clickable - not just the linked text

app/views/vacancy_submissions/_search_form.html.haml
```diff
@@ -1,9 +1,6 @@
...
-    .col-auto
-      = f.text_field :q, value: params[:q], placeholder: 'Search by program name...', class: 'form-control form-control-sm'
-    .col-auto
-      - status_options = [['Queue (Awaiting + Changes Requested)', 'queue'], ['All Statuses', 'all'], ['Awaiting Approval', 'awaiting_approval'], ['Return / Changes Requested', 'return_changes_requested'], ['Active', 'active']]
-      = f.select :status, status_options, { selected: params[:status].presence || 'queue' }, class: 'form-select form-select-sm'
-    .col-auto
-      = f.submit 'Search', class: 'btn btn-sm btn-secondary'
+= form_tag(vacancy_submissions_path, method: :get) do
+  .input-group
+    - status_options = [['Queue (Awaiting + Changes Requested)', 'queue'], ['All Statuses', 'all'], ['Awaiting Approval', 'awaiting_approval'], ['Return / Changes Requested', 'return_changes_requested'], ['Active', 'active']]
```
dtgreiner, line 3: This feels like we should be pulling it from VacancySubmissionsHelper::STATUS_DISPLAY and adding `queue` and `all`. Maybe something like this
```suggestion
    - status_options = [['Queue (Awaiting + Changes Requested)', 'queue'], ['All Statuses', 'all']] + VacancySubmissionsHelper::STATUS_DISPLAY.map { |k, v| [v, k] }
```

dtgreiner (approved):
Looks good! A couple comments

app/views/vacancy_submissions/index.html.haml
eanders: Looks like we only use it in one other place in the app.  
Good point about the cursor.  Given the small use of it, I'm just removing the `cursor: pointer` declaration.  So the row will still get a subtle highlight, but the cursor only becomes a pointer if over a link.
Thanks!

app/views/vacancy_submissions/_search_form.html.haml
eanders: Agreed, good catch.  Thanks!

PR #1129, opened 2026-07-11
9379 de-identified upload enhancement

ttoomey:
## _Merging this PR_

### Changes

[See Issue](https://github.com/open-path/Green-River/issues/9379#issuecomment-4942271389)

- **Cross-agency ID collisions.** Client identifiers are globally unique. When an uploaded ID already belonged to another agency, the save failed the uniqueness validation silently, so the client looked ineligible and stayed on the wrong agency. The importer now looks up by ID, skips those rows, and reports the skipped IDs to the user.
- **Atomic availability reset.** The "update availability" reset (mark all agency clients unavailable, then re-activate those in the file) now runs in a single transaction, so a mid-import failure rolls back instead of stranding clients as unavailable or leaving stale ones active.
- **Upload form.** Agency and availability are now explicit required selections (availability is a Yes/No radio; agency defaults to blank rather than the uploader's own agency). The agency list is restricted to agencies the user may edit, mirroring single-record permissions. Errors re-render preserving only submitted values.
- **Error reporting.** Internal failures are reported to Sentry instead of being swallowed; user-correctable data problems still render in the per-row problems table.

## Type of change
Bug fix, New feature

### DIFF
--- a/app/models/deidentified_clients_xlsx.rb
+++ b/app/models/deidentified_clients_xlsx.rb
@@ -45,14 +45,11 @@ class DeidentifiedClientsXlsx < ApplicationRecord
     return unless valid_header?
 
     # Internal (non-user) errors are buffered here and reported to Sentry only after the
-    # transaction commits — reporting mid-transaction would flag rows for a run that may still
-    # roll back and persist nothing. On rollback the block below raises out, this flush is
-    # skipped, and the propagating exception is what reaches Sentry instead.
-    deferred_reports = []
+    # transaction commits
+    sentry_queue = []
 
     # Availability is reset up front, so the whole roster must be atomic: a non-recoverable
-    # failure partway through has to roll that reset back rather than leave every client in
-    # the agency stranded as unavailable.
+    # otherwise a failure could leave all clients in the agency unavailable.
     DeidentifiedClient.transaction do
       DeidentifiedClient.where(agency: agency).update_all(available: false) if @update_availability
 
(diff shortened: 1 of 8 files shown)


app/models/deidentified_clients_xlsx.rb
```diff
@@ -38,44 +38,86 @@ def valid_header?
...
+
+      @xlsx.each_with_index do |raw, index|
+        next if skip?(raw, index)
+
+        row = Hash[file_attributes.keys.zip(raw)]
+
+        # A Home-base ID is globally unique, so look the client up by ID alone If it's already
+        # live under a *different* agency skip and report it
+        client = DeidentifiedClient.find_by(client_identifier: row[:client_identifier]) ||
+          DeidentifiedClient.new(agency: agency, client_identifier: row[:client_identifier])
```
eanders, line 64: Do we enforce that `client_identifier` is globally unique?  I think there's a high likelihood that two agencies might have overlapping `client_identifier`s representing different clients.  


Answering my own question, it looks like we do enforce it with a validation only:
```
validates :client_identifier, uniqueness: true
```

So I guess this is ok and an improvement.

app/models/deidentified_clients_xlsx.rb
```diff
@@ -38,44 +38,86 @@ def valid_header?
...
+        if client.persisted? && client.agency_id != agency&.id
+          @skipped_identifiers << row[:client_identifier] unless @skipped_identifiers.include?(row[:client_identifier])
+          next
+        end
+
+        @clients << client
+        cleaned = begin
+          clean_row(client, row)
+        rescue StandardError => e
+          # clean_row's helpers attach a field-level error before raising. Tthose are expected,
```
eanders:
```suggestion
          # clean_row's helpers attach a field-level error before raising. Those are expected,
```

eanders (approved):
Thanks @ttoomey, just one typo, then I think this is good.

PR #1135, opened 2026-07-16
[#9393] Vacancy Submission Updates

eanders:
## _Merging this PR_

## Description

* pivot to building pick list with fancy address (so you can type-ahead with building name or address)
* Expose a note when working with a confidential program that you should not use a real address
* Fix link to sub-program
* Correctly add and increment weighting rules to new vouchers
* Notifications for submission, changes requested, and resubmission
* Markdown helper, optimistic cleanup of two existing markdown uses

## Type of change

* Bug fix
* New feature


app/mailers/vacancy_submissions_mailer.rb
eanders, line 1: I think this is probably how we should refactor all existing mailers eventually.  The "view" side is basically `= @body` which would mean we could eventually streamline the views down to two templates, one for HTML and one for plain text.  Then each part of the message can be translated appropriately and generated by the mailer class.

app/models/vacancy_submissions/notifier.rb
```diff
@@ -0,0 +1,43 @@
...
+      return if submitter.blank?
+
+      VacancySubmissionsMailer.changes_requested(@vacancy_submission, submitter).deliver_later
+    end
+
+    private
+
+    def reviewers
+      User.vacancy_reviewers
+    end
```
eanders, line 41: This might need to change later, maybe the contacts should be assigned to a program, or building or something, but for now, we're alerting *everyone* who can approve it when there's a new or resubmitted vacancy.

app/helpers/markdown_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+
+# frozen_string_literal: true
+
+module MarkdownHelper
+  # Render a Markdown string to HTML, enabling HTML tags and links.
+  def render_markdown(text)
+    return ''.html_safe if text.blank?
+
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true)
+    markdown.render(text).html_safe
```
dtgreiner, line 25: Should we include a note on this method/file that anything passed in will be flagged as html_safe? It doesn't appear to be an issue with how it's currently being used, but if it is used in the future for user-entered text, it could open up a vulnerability

app/views/vacancy_submissions/_vacancy.html.haml
```diff
@@ -2,4 +2,4 @@
 - if is_voucher
   = render 'vacancy_submissions/vacancy_vouchers', unit_list: unit_list
 - else
-  = render 'vacancy_submissions/vacancy_units', unit_list: unit_list
+  = render 'vacancy_submissions/vacancy_units', unit_list: unit_list, default_building_id: sub_program&.building_id, confidential: sub_program&.confidential?
```
dtgreiner, line 5:
```suggestion
  = render 'vacancy_submissions/vacancy_units', unit_list: unit_list, default_building_id: sub_program&.building_id, confidential: sub_program && (sub_program.confidential? || sub_program.program.confidential?)
```

app/mailers/vacancy_submissions_mailer.rb
dtgreiner: This does feel much nicer

dtgreiner (approved):
@eanders - just a couple notes. The confidential project display is currently only showing up when the sub_project specifically is marked as confidential. Unless I'm misunderstanding the confidentiality feature, that should probably be addressed, but otherwise LGTM

PR #1149, opened 2026-08-10
fix: 9487 don't de-list de-identified clients on parse-error

ttoomey:
## _Merging this PR_

## Description
The up-front availability reset was never undone for skipped rows, so one bad cell silently made a client who was in the file ineligible for matching. Failed rows now restore their prior availability, making them a true no-op.

## Type of change
Bug fix

### DIFF
--- a/app/models/deidentified_clients_xlsx.rb
+++ b/app/models/deidentified_clients_xlsx.rb
@@ -48,9 +48,14 @@ class DeidentifiedClientsXlsx < ApplicationRecord
     # transaction commits
     sentry_queue = []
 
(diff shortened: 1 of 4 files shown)


eanders:
@ttoomey can you re-target `release-104`?  107 probably won't be deployed for weeks yet, given that this is in-use and causing an issue, I think I'd like to get it out sooner.

eanders (approved):

ttoomey:
I can retarget. Just to clarify this isn't a new bug, it has probably been there for years; I just discovered it.

eanders:
Given that we just fixed another related bug, it feels like this might as well go out sooner, the feature is being used.

PR #1151, opened 2026-08-13
Markdown and Sentry fixes

eanders:
## _Merging this PR_

## Description

* Fixes a bug where the markdown rendered wasn't included on a page that used it (checked, all other instances appear to have it.)
* Ported the recent sentry changes from the warehouse after realizing sentry notifications were being dropped silently.

## Type of change

* Bug fix

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -137,7 +137,7 @@ gem 'browser'
 gem 'net-http'
 gem 'ajax_modal_rails', '~> 1.0'
 
-gem 'sentry-rails', '~> 5.5'
+gem 'sentry-rails', '~> 6'
 gem 'warning'
 
 # Metrics
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -246,7 +246,7 @@ GEM
     caxlsx_rails (0.6.4)
       actionpack (>= 3.1)
       caxlsx (>= 3.0)
-    cgi (0.5.1)
+    cgi (0.5.2)
     childprocess (5.1.0)
       logger (~> 1.5)
     chronic (0.10.2)
@@ -260,7 +260,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.12.2)
-    concurrent-ruby (1.3.7)
+    concurrent-ruby (1.3.8)
     connection_pool (2.5.5)
     crass (1.0.7)
     csv (3.3.5)
@@ -323,7 +323,7 @@ GEM
       dry-inflector (~> 1.0)
       dry-logic (~> 1.4)
       zeitwerk (~> 2.6)
-    erb (6.0.4)
+    erb (6.0.7)
     erubi (1.13.1)
     erubis (2.7.0)
     excon (0.112.0)
@@ -365,12 +365,12 @@ GEM
       nokogiri (>= 1.6.0)
       ruby_parser (~> 3.5)
     htmlentities (4.4.2)
-    i18n (1.14.8)
+    i18n (1.15.2)
       concurrent-ruby (~> 1.0)
     ice_nine (0.11.2)
     iniparse (1.5.0)
-    io-console (0.8.2)
-    irb (1.17.0)
+    io-console (0.9.2)
+    irb (1.18.0)
       pp (>= 0.6.0)
       prism (>= 1.3.0)
       rdoc (>= 4.0.0)
@@ -475,7 +475,7 @@ GEM
       racc
     pg (1.6.3)
     pg (1.6.3-arm64-darwin)
-    pp (0.6.3)
+    pp (0.6.4)
       prettyprint
     pretender (0.6.0)
       actionpack (>= 7.1)
@@ -492,15 +492,12 @@ GEM
       pry (>= 0.13, < 0.17)
     pry-rails (0.3.11)
       pry (>= 0.13.0)
-    psych (5.3.1)
-      date
-      stringio
     public_suffix (7.0.5)
     puma (8.0.2)
       nio4r (~> 2.0)
     pwned (2.4.1)
     racc (1.8.1)
-    rack (3.1.21)
+    rack (3.1.22)
     rack-attack (6.8.0)
       rack (>= 1.0, < 4)
     rack-mini-profiler (4.0.1)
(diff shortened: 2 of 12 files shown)


ttoomey (commented):
Thanks - Can you take a look at these findings from AI review?

---

## 1. High-Level Summary

Two unrelated fixes. First, `NonHmisAssessmentsController` now declares `helper MarkdownHelper`, restoring the TC-HAT assessment form pages that render a field's `description:` as Markdown — `include_all_helpers` is off (`config/application.rb:91`), so the helper was never mixed in and `render_markdown` raised `NoMethodError`. Second, the Sentry `before_send` hook moves from an inline lambda into `SentryEventFilter`, which scrubs individual event fields in place and returns the `Sentry::ErrorEvent` instead of a filtered `Hash`.

## 2. File-by-File Review

**1. `[critical]` `app/controllers/non_hmis_assessments_controller.rb:10`** — The fix covers only `NonHmisAssessmentsController`. `NonHmisClientsController#current_assessment_limited` (`app/controllers/non_hmis_clients_controller.rb:162`) renders the same partials and still has no `MarkdownHelper`, so it raises `NoMethodError: undefined method 'render_markdown'`. Two paths reach it:

- **Pathways V3.** `identified_clients/current_assessment_limited.haml:6` renders `_identified_pathways_version_three.haml:3` → `_common_pathways_version_three_questions.haml:3-4`, which renders `options[:partial]` before any `disabled` branch → `_pathways_version_three_preamble.haml:20` calls `render_markdown`. The preamble is registered in `pathways_form_fields` (`pathways_version_three_calculations.rb:872-874`), so this fires for any V3 assessment whose title is `'Pathways 2021'`.
- **TC HAT.** `_identified_tc_hat.haml:2` → `_common_tc_hat_questions.haml:67` → `_common_description_display.haml:3`. Line 67 is reached even when `disabled` is true: the guard on line 52 tests `@assessment.respond_to?(field_name)`, which is false for the underscore-prefixed pseudo-field `_tc_hat_placed_on_list_note` (`tc_hat_calculations.rb:244-248`; no matching column in `db/schema.rb`), so control falls through to the `elsif field_name.starts_with?('_')` branch.

The route exists for both client types (`config/routes.rb:300` and `:311`) and is linked from `non_hmis_assessments/_index.haml:56` and `matches/_client_match_attributes.haml:9`. The fix is to add `helper MarkdownHelper` to `NonHmisClientsController`; because `_helpers` is an inherited `class_attribute`, `IdentifiedClientsController` and `DeidentifiedClientsController` both pick it up. `ImportedClientsController` has no assessment views and is unaffected.

**2. `[warning]` `lib/util/sentry_event_filter.rb:16-20` — breadcrumb data is no longer scrubbed (regression).** The old code was `filter.filter(event.to_hash)`, and `Sentry::Event#to_hash` (`event.rb:121-125`) includes `data[:breadcrumbs] = breadcrumbs.to_hash`, so the whole serialized breadcrumb list passed through `ParameterFilter`. The new `call` scrubs only `tags`, `extra`, `contexts`, `user`, and four `request` sub-fields — `event.breadcrumbs` is untouched. With `sentry-rails` installed, breadcrumbs carry ActiveSupport instrumentation payloads (SQL, controller params), so values that used to be redacted now leave the process in the clear. This is the one place where returning the event object narrows coverage rather than widening it, and it's the sort of thing the whole `sanitize_fields` mechanism exists to prevent.

`Sentry::BreadcrumbBuffer` is `Enumerable` (`breadcrumb_buffer.rb:8,41`) and `Sentry::Breadcrumb#data` is a plain accessor (`breadcrumb.rb:11`), so the addition is a one-liner in the same in-place style as the rest of `call`, guarded on `data.is_a?(Hash)` the way `request.data` already is. The spec file has no breadcrumb case either.

**3. `[warning]` `lib/util/sentry_event_filter.rb:12-15`** — The comment describes sentry-ruby 6.x behavior, but `Gemfile.lock:639` pins `sentry-ruby (5.28.1)`. Reading the installed gem (not 5.27.0): `Client#send_event` (`client.rb:220-235`) has `case event ... when Hash then log_debug("Returning a Hash from before_send is deprecated and will be removed in the next major version.")` — no discard, no `record_lost_event`. `Transport#envelope_from_event` (`transport.rb:114-159`) then resolves `item_type = event_payload[:type] || event_payload["type"]`, which works fine on a Hash, so the envelope was well-formed and the event was sent. Only the `dynamic_sampling_context` (line 127) and `attachments` (line 156) blocks were skipped, both guarded on `event.is_a?(Event)`.

So two clauses of the comment are wrong for the version this app runs: nothing was "malformed," and nothing was discarded. The outright discard (`Discarded event because before_send didn't return a Sentry::ErrorEvent object`) exists only in 6.x, which is what the warehouse runs. Consequences: the comment will mislead the next reader, and — more importantly — the stated reason for the missing Sentry notifications does not apply on 5.28.1, so whatever is actually dropping events here is probably still unfixed. Worth establishing that before closing the ticket.

**4. `[warning]` `spec/requests/non_hmis_assessments_spec.rb:145-163`** — The new spec covers one of the two views that call `render_markdown` and one of the two affected controllers. Neither `_pathways_version_three_preamble.haml:20` nor the `current_assessment_limited` action is exercised, so the suite passes while the bug in finding 1 is live.

`config/initializers/sentry.rb` and the field-scrubbing in `SentryEventFilter#call` itself: no further issues. Returning the event object is correct on both 5.x and 6.x, and it restores the `dynamic_sampling_context` and attachment handling that the `Hash` return skipped — the change is right, it just needs the breadcrumb case to reach parity with what `to_hash` covered.

## 3. Recommended Next Steps

1. Add `helper MarkdownHelper` to `NonHmisClientsController` (finding 1).
2. Add breadcrumb scrubbing to `SentryEventFilter#call`, plus a spec case (finding 2) — this one is a live PII exposure, not just a robustness nit.
3. Extend the request spec to cover `current_assessment_limited` for both a TC HAT and a `'Pathways 2021'` V3 assessment (finding 4).
4. Correct the comment to describe 5.x behavior, and separately determine what was actually dropping Sentry events on 5.28.1 (finding 3).

eanders:
Thanks @ttoomey.  Funny how one AI will swear up and down it has it all covered and another finds holes.  
I've pushed a change that:

1. Bumps sentry to 6.7 - aligns with the warehouse
2. Includes a rake task to test the key sentry integration points (ran locally, [looks good](https://green-river.sentry.io/issues/7670778899/?project=4504017585569792&query=is%3Aunresolved&referrer=issue-stream) for example.
3. Scrubs breadcrumb events
4. Adds test coverage as noted above
5. Moves MarkdownHelper to ApplicationController.  This popped up in two places today, felt like there's no reason to keep fighting it in the future

Let me know if I missed anything, and thanks again!

ttoomey (approved):
Looks good, thanks!

